PR #1, opened 2026-08-31
[Design] Enrolment and payment flow

Abhigna1975:
Design for the payment gateway, under `design/payment-gateway/`. **No application code changes** — this is the flow settled before it gets built.

Canvas: **https://claude.ai/code/artifact/96c5f202-1c3f-4cea-ae2f-4f40f747ebf2**

Built against this repo's own palette (`src/styles/colors.css` — navy `rgb(3,8,76)`, `#146389`, `#ff9800`) and Montserrat. Nothing new invented.

| Board | |
|---|---|
| 01 | Course card → checkout → confirmation, mobile-first |
| 02 | The referral field, and why it's built this way |
| 03 | The two open commercial decisions, shown both ways |
| 04 | What this needs from the app that already exists |

## It extends the enquiry form rather than duplicating it

`EnquiryForm.jsx` already collects full name, mobile, email and experience — the same fields a checkout needs, already validated, already posting to `trcabe.onrender.com`. The checkout reuses those atoms so the app doesn't grow a second form language.

**"Or talk to a counsellor first" stays on the course card.** The enquiry funnel is what converts today. Payment is added beside it, not in front of it — swapping a working funnel for an untested one isn't a design decision anyone should make quietly.

## The referral field is the part worth reviewing

The default pattern — "Have a coupon?" behind a disclosure link — loses the referral on most visits. A student who *was* referred but never opens that link pays in full and records as organic. Money collected, attribution lost, and nothing anywhere reports an error.

So it's **always visible**, pre-filled from `?ref=…` on landing, persisted while they browse, and confirmed in green **before** the pay button. A code that silently fails to apply is worse than no field — it looks handled and isn't.

It should also ride into Razorpay's `notes` on order creation, so it lands in the payment record and can be queried instead of counted by hand.

## Two decisions still open

1. **Full fee online, or a booking deposit?** A deposit is far easier to say yes to, but moves most of the fee offline.
2. **Do prices go public?** There's no pricing anywhere on the site today.

Board 03 shows both, both ways. Either answer is a config change on this layout, not a redesign — the balance row appears when the amount is under the fee, and the price on the course card is one boolean.

## Build notes worth agreeing before code

- **The amount comes from the server, never the form.** A price posted from the browser is a price the browser can change.
- **Verify the Razorpay signature server-side.** A redirect alone can be forged.
- **Idempotency** — a double-tap must not create two orders.
- **Not designed yet:** payment failure and retry, refunds/cancellation, desktop. Failure matters most — on a payment this size it decides whether the student retries or rings a competitor. Quick to add once the two decisions above land.


nikshepkulli (approved):
Design approved — this is the right shape for the flow, and the review-worthy calls are all correct:

- **Extends the enquiry form rather than duplicating it** — reusing the validated `EnquiryForm.jsx` atoms so the app doesn't grow a second form language. Good.
- **Payment beside the counsellor funnel, not in front of it** — right. The enquiry funnel is what converts today; don't swap a working funnel for an untested one quietly.
- **The always-visible referral field, pre-filled from `?ref=`, confirmed green before Pay, and ridden into Razorpay `notes`** — this is the part that matters most for us. A silently-failing coupon-behind-a-disclosure loses attribution on most visits: money collected, attribution lost, no error. Always-visible + in `notes` is exactly how the rev-share stays measurable.
- **Build notes are all correct and non-negotiable:** amount from the server never the form; verify the Razorpay signature server-side; idempotency on double-tap. Please keep these as hard requirements when it's built.
- **Payment-failure/retry is the right thing to design next** — on a ₹35k payment it decides whether the student retries or rings a competitor.

**The two open commercial decisions (full fee vs deposit; prices public) are founder calls — I'm settling them now and will post the answers here so the config choices on board 03 can be locked.**

nikshepkulli:
**Both commercial decisions are settled — board 03 configs lock:**

1. **Full ₹35k online** (no deposit). Rev-share is online-only, so the whole fee flows through Razorpay → we're paid on 100% and every enrolment is attributable by referral code. The counsellor funnel stays beside it for anyone not ready to pay. → On the layout: **no balance row** (amount always equals the full fee).

2. **Prices go public** — show the price on the course card. → The price boolean = **on**.

So neither config is 'both ways' anymore — build to: full amount, price shown. Everything else in the flow (server-side amount, signature verify, idempotency, referral in `notes`) stays as you specced. Payment-failure/retry is the right next design once this is in code.

Approved and ready to move to build.

nikshepkulli:
Scope note from #2: the current enquiry backend (`trcabe.onrender.com`) will be **consolidated into this payment build**, not maintained separately — one backend for enquiry + enrolment. It's being revived off Render's free tier short-term (#2) to stop dropping leads now; the durable replacement lands here. Keep the server-side Razorpay order + signature verify + idempotency as specced — the enquiry POST folds in alongside.

Issue #6, opened 2026-09-01
[P1] Adopt design tokens: one typeface, four radii


Abhigna1975:
`design/tokens.css` and `design/ui-template.html` are open for review as PR #14. This ticket adopts them once it merges.

Right now the same values are re-typed per file and have drifted:

**Font.** Montserrat is loaded in `src/styles/fonts.css` but only applies through the `.ff-MS` class. The actual body font is `Segoe UI`, forced in `src/index.css:8` with `!important`. The navbar and footer use **Tilt Neon** (`Navbar.css:20,39`, `FooterComponent.css:22,44`). **Dancing Script** is loaded in `index.html` and never used anywhere. So the brand typeface is largely not rendering.

**Radius.** Ten different values ship today: `0`, `3px`, `4px`, `5px`, `10px`, `.3rem`, `30px`, `50%`, `100%`.

### What to build

- [ ] `import "../design/tokens.css"` in `src/main.jsx`, above every other stylesheet.
- [ ] Montserrat becomes the one face. Remove the `Segoe UI` `!important` override, replace Tilt Neon in the navbar and footer, delete the unused Dancing Script link from `index.html`.
- [ ] Replace hardcoded colours with the variables. `src/styles/colors.css` becomes a thin layer over the tokens, or goes away.
- [ ] Collapse the ten radii to the four in the template: chips 4, buttons and inputs 8, cards and modals 12, tags pill.
- [ ] Section gaps and gutters onto the spacing scale.

### Done when

`grep -rE "#[0-9a-fA-F]{6}|rgb\(" src/` returns nothing outside `tokens.css`, and one typeface renders across the site.

Issue #5, opened 2026-09-01
[P1] Navbar hamburger is a dead control at 600px and above


Abhigna1975:
`src/components/molecules/Navbar/Navbar.jsx` renders the hamburger `IconButton` with `sx={{ display: { sm: 'block', xs: 'block' } }}` — MUI breakpoints being min-width, `sm: 'block'` means it stays visible at **every width from 600px up**.

The Drawer it opens is `sx={{ display: { xs: 'block', sm: 'none' } }}` — hidden from 600px up.

So from 600px onward there is a hamburger button that does nothing, sitting next to the inline nav links, which appear at the same breakpoint. Two navigations, one of them dead.

The icon colour compounds it: `sx={{ color: { sm: 'black', xs: 'black', lg: 'white' } }}` puts a black icon on the navy AppBar between 600px and 1200px.

### What to build

- [ ] Hamburger visible below `md` only, inline nav links `md` and up. One nav visible at a time, at every width.
- [ ] Drawer breakpoint matched to the hamburger's, so the button always opens something.
- [ ] Icon colour fixed — it sits on navy, so it is white.
- [ ] Drawer rows at least 44px tall (section 7 of `design/ui-template.html`).

### Done when

375 / 768 / 1024 / 1440 each show exactly one working navigation, and no visible control is inert.

Issue #4, opened 2026-09-01
[P0] Landing page ships ~7MB of images; one PNG is 3.8MB


Abhigna1975:
The landing page ships roughly **7MB**, measured on the live deploy:

| Asset | Size |
|---|---|
| `sakshi.png` | **3.8 MB** |
| `prajwala.png` | 788 KB |
| `ashish.png` | 631 KB |
| `subhash.png` | 498 KB |
| `naseema.png` | 482 KB |
| `sujith.png` | 482 KB |
| other logos | ~240 KB |
| JS bundle | 729 KB (228 KB gzipped) |

On a mobile connection that is several seconds of blank screen. Most of the traffic to a Bangalore coaching site is on a phone, and a visitor who leaves before the hero paints was never going to see the courses, the placements or the enquiry button.

### What to build

- [ ] Compress every image in `src/assets/`. Export at 2× the size it actually displays at, WebP with a JPG fallback. Target the whole page under **1MB**.
- [ ] Set explicit `width`/`height` (or `aspect-ratio`) on every image so nothing reflows as they load.
- [ ] `loading="lazy"` on everything below the fold — reviews, placements, clients.
- [ ] Drop the non-web files that don't belong in `src/`: `src/assets/reviews/Rest coder.zip` (8.9MB) and the four `.ai` Illustrator files (1.2MB each).

### Done when

Total page weight is under 1MB and no single image is over 200KB.

Issue #3, opened 2026-09-01
[P0] Batch dates are hardcoded to 24-05-2025


Abhigna1975:
The Batches section — the one part of the site that signals "we are actually running classes" — advertises batches starting **24-05-2025**. That is 15 months in the past.

Every entry in `src/components/organism/Batches/batches.js` is hardcoded to `date:"24-05-2025"`. Anyone who reads that date concludes the academy is dormant, and they're right to.

### What to build

- [ ] Update the batch dates to the real upcoming ones (needs the current schedule from Uday).
- [ ] A date in the past must never render as an upcoming batch. Render it as **Batch closed** / **Batch full** using the muted tag in section 5 of `design/ui-template.html`, with the CTA switching to "Join the waitlist".
- [ ] Show the next batch date on the course card too (the `Next batch · 15 Sep` tag in the template) — it's the single strongest reason to enquire today rather than next month.
- [ ] Keep all dates in the one data file. No dates in components.

### Done when

No past date is presented as upcoming anywhere on the site, and the batch date is visible from the course card without scrolling to the Batches section.

nikshepkulli:
Follow-up filed: **#15** makes the batch dates admin-controlled so RCA updates them themselves without a redeploy — the durable fix so they never go stale again. #3 stays the immediate P0 (stop advertising the 15-month-old date now); #15 is the P1 self-serve layer on top. Assigned to @Pavankulli.

PR #17, opened 2026-09-01
Fix enquiry form silently losing leads on submit failure

### LINKED ISSUE #2: [P0] Enquiry form silently loses every lead
Every enquiry submitted right now is lost, and nobody finds out.

`src/components/forms/Enquiry Form/EnquiryForm.jsx:57-70` posts to `https://trcabe.onrender.com/enquiries/create/enquiry`. That host did not respond within 70 seconds when checked on 31 Aug 2026. The `catch` block does nothing but `console.log(err)`, and the `useEffect` below it calls `notify()` and `closeModal()` regardless — so the person sees **"Thanks {name}. We will reach you soon"**, the modal closes, and their details go nowhere.

The enquiry form is the only place a lead is captured on this site. At a ₹35,000 course fee, each one of these is a ₹35,000 enrolment that never gets a call back.

### What to build

Follow section 4 of `design/ui-template.html`.

- [ ] Await the POST. Only show success on a `2xx`.
- [ ] Failure → keep the modal open, keep what they typed, and show the error banner: *"We couldn't send that just now. Message us on WhatsApp and we'll pick it up straight away."* with the existing `wa.me/918073762257` link as a working fallback so the lead still reaches someone.
- [ ] In flight → button shows the loading state (spinner, label `Sending…`, disabled) so it can't be double-submitted.
- [ ] Success → confirmation naming the number we'll call: *"Enquiry received. We'll call you on {mobile} within one working day."*
- [ ] Add a submit timeout (~10s) so a cold or dead backend fails visibly instead of hanging.

### Note

Whether the `trcabe.onrender.com` backend gets revived or replaced is Nik's call and not part of this ticket. This ticket makes the UI honest either way — the WhatsApp fallback means leads keep arriving even while the backend is down.

### Done when

Backend up → lead saved, success shown. Backend down → error shown, modal stays open, WhatsApp path offered. No path shows success without a `2xx`.

Pavankulli:
## Summary
- Await the enquiry POST and only show success on a `2xx`; previously the form showed a fake success toast and closed regardless of the request's outcome, silently dropping every lead whenever the backend failed.
- On failure, keep the modal open with the user's typed data intact and show an error banner with a working WhatsApp fallback link (`wa.me/918073762257`) so the lead still reaches someone.
- Submit button shows a loading spinner + "Sending…" and disables itself while in flight, preventing double-submits.
- Add a ~10s request timeout so a cold/dead backend fails visibly instead of hanging.
- Success toast now reads "Enquiry received. We'll call you on {mobile} within one working day."
- Polish the enquiry modal UI (responsive sizing, shadow, header, buttons, error banner) and the overlay backdrop.
- Fix two pre-existing bugs surfaced while testing: text inputs/textarea rendering with unreadable dark backgrounds on browsers/OS in dark mode, and the modal rendering underneath the navbar due to a z-index clash with MUI's `AppBar`.

Fixes #2

## Test plan
- [x] `npm run lint` shows no new errors (pre-existing repo-wide lint errors unaffected)
- [x] Manually verify in a browser: submit with backend down → spinner → error banner + WhatsApp link, modal stays open, data preserved, no double-submit
- [x] Manually verify success path once a live/mock backend returns 2xx: success toast names the mobile number, modal closes
- [x] Verify modal renders above the navbar and matches expected styling in both light and dark browser themes

### DIFF
--- a/src/components/forms/Enquiry Form/EnquiryForm.jsx
+++ b/src/components/forms/Enquiry Form/EnquiryForm.jsx
@@ -10,7 +10,8 @@ import { useAuth } from '../../../App';
 import { regex } from '../../../regex/regex';
 import axios from 'axios';
 
-const WHATSAPP_FALLBACK_URL="https://wa.me/918073762257"
+const WHATSAPP_FALLBACK_NUMBERS=["918073762257","919110424403"]
+const WHATSAPP_PREFILL=encodeURIComponent("Hi Rest Coder Academy, I'd like to enquire about a course.")
 const SUBMIT_TIMEOUT_MS=10000
 
 function EnquiryForm() {
@@ -218,14 +219,19 @@ const handleSubmit = async (e)=>
               variant="body2"
               text="We couldn't send that just now. Message us on WhatsApp and we'll pick it up straight away."
             />
-            <a
-              className="enquiry-whatsapp-fallback"
-              href={WHATSAPP_FALLBACK_URL}
-              target="_blank"
-              rel="noopener noreferrer"
-            >
-              Message us on WhatsApp
-            </a>
+            <div className="enquiry-whatsapp-links">
+              {WHATSAPP_FALLBACK_NUMBERS.map((num) => (
+                <a
+                  key={num}
+                  className="enquiry-whatsapp-fallback"
+                  href={`https://wa.me/${num}?text=${WHATSAPP_PREFILL}`}
+                  target="_blank"
+                  rel="noopener noreferrer"
+                >
+                  WhatsApp {num.replace(/^91/, "").replace(/(\d{5})(\d{5})/, "$1 $2")}
+                </a>
+              ))}
+            </div>
           </Box>
         )}
         <Box className="enquiry-field-button">
--- a/src/components/forms/Enquiry Form/EnquiryForm.jsx
+++ b/src/components/forms/Enquiry Form/EnquiryForm.jsx
@@ -64,18 +64,13 @@ const handleSubmit = async (e)=>
     try
     {
       // https://trcabe.onrender.com/enquiries/get/enquiries
-      let response=await axios.post('https://trcabe.onrender.com/enquiries/create/enquiry',enquiryData,{timeout:SUBMIT_TIMEOUT_MS})
+      // axios only resolves for 2xx responses by default; anything else (and network
+      // errors/timeouts) rejects and is handled in the catch block below.
+      await axios.post('https://trcabe.onrender.com/enquiries/create/enquiry',enquiryData,{timeout:SUBMIT_TIMEOUT_MS})
       // await axios.post('http://localhost:4005/enquiries/create/enquiry',enquiryData)
 
-      if(response.status>=200 && response.status<300)
-      {
-        notify(`Enquiry received. We'll call you on ${enquiryData.mobile} within one working day.`)
-        closeModal()
-      }
-      else
-      {
-        setSubmitFailed(true)
-      }
+      notify(`Enquiry received. We'll call you on ${enquiryData.mobile} within one working day.`)
+      closeModal()
     }
     catch(err)
     {
@@ -158,7 +153,7 @@ const handleSubmit = async (e)=>
           </ButtonComponent>
       </Box>
 
-      <Box className="enquiry-fields" p={2}>
+      <Box className="enquiry-fields">
         <Box className="enquiry-field">
           <InputBoxComponent
             value={enquiryData.fullname}
--- a/src/components/forms/Enquiry Form/EnquiryForm.jsx
+++ b/src/components/forms/Enquiry Form/EnquiryForm.jsx
@@ -154,7 +154,7 @@ const handleSubmit = async (e)=>
        paddingY={1}
             onBtnClick={closeModal}
           >
-           X
+           ×
           </ButtonComponent>
       </Box>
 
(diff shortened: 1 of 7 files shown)


src/components/forms/Enquiry Form/EnquiryForm.jsx
```diff
@@ -38,38 +42,46 @@ let changeDropDown=(value)=>
     setenquiryData({...enquiryData,experience: value.label})
 }
     
-const handleSubmit = (e)=>
+const handleSubmit = async (e)=>
```
Pavankulli, line 44: Fixed in a2c738c — added an isSubmitting re-entrancy guard at the top of handleSubmit.

src/components/forms/Enquiry Form/EnquiryForm.jsx
```diff
@@ -1,5 +1,5 @@
-import { Box, Typography,Grid, Paper } from '@mui/material';
-import React, { useEffect, useState } from 'react'
+import { Box, Typography,Grid, Paper, CircularProgress } from '@mui/material';
```
Pavankulli: Fixed in a2c738c — removed the unused Typography/Grid/Paper imports.

src/components/forms/Enquiry Form/EnquiryForm.jsx
```diff
@@ -17,10 +17,14 @@ function EnquiryForm() {
   let [enquiryData,setenquiryData]=useState({fullname:"",mobile:"",email:"",experience:"",message:""})
   // const { register, handleSubmit, watch, formState: { errors } , control, reset} = useForm();
   let [enquiryErrors,setenquiryErrors]=useState({})
-  let [issubmit,setissubmit]=useState(false)
+  let [isSubmitting,setIsSubmitting]=useState(false)
+  let [submitFailed,setSubmitFailed]=useState(false)
 
   let {closeModal,notify}=useAuth()
 
+  const WHATSAPP_FALLBACK_URL="https://wa.me/918073762257"
```
Pavankulli: Fixed in a2c738c — hoisted WHATSAPP_FALLBACK_URL and SUBMIT_TIMEOUT_MS to module scope.

Pavankulli:
## Code review

Core fix looks solid: replacing the `useEffect`-driven auto-submit with an explicit `await sendEnquiry()` gated on validation, plus the `submitFailed`/WhatsApp-fallback UX and `isSubmitting` spinner, correctly closes the original "lead lost on submit failure" bug (the old code could also auto-resubmit on a later unrelated render once `issubmit` was stuck `true`).

**Correctness**

1. `EnquiryForm.jsx` — `handleSubmit` (~line 45) has no re-entrancy guard against double submission; it relies solely on `disabled={isSubmitting}` on the Submit button rather than checking `isSubmitting` itself. Since this PR is specifically about not losing/duplicating leads, a defensive `if(isSubmitting) return` at the top of `handleSubmit` is cheap insurance.
2. `EnquiryForm.jsx` — the catch block (~line 75) treats every rejection the same: network errors, the new 10s axios timeout, and any 4xx business-logic rejection from the backend all show the identical "couldn't send, message us on WhatsApp" fallback. Two concrete issues: (a) on a client-side timeout the POST may already have been received/processed server-side, so the WhatsApp fallback risks creating a *duplicate* lead rather than preventing a lost one; (b) a genuine validation/4xx rejection gets misrepresented to the user as a connectivity problem.

**Cleanup**

3. `EnquiryForm.jsx:1` — `Typography`, `Grid`, `Paper` are imported but never used anywhere in the file (only the custom `TypoGraphyComponent` renders text). This exact line is touched by the "clean up dead code from self-review" commit, so it's a natural place to also drop these — ESLint flags all three as `no-unused-vars`.
4. `EnquiryForm.jsx:25-26` — `WHATSAPP_FALLBACK_URL` and `SUBMIT_TIMEOUT_MS` don't depend on props/state but are declared inside the component body, so they're re-created every render. Worth hoisting to module scope.
5. `App.css:79` — `.modal-enquiry-form` sits inside `.ReactModal__Overlay` (already `z-index: 1300`), so the content div's own `z-index: 1301` is a no-op for paint order — harmless but confusing.
6. `EnquiryForm.jsx` close button (~line 152) — glyph changed from the letter `X` to `×` (U+00D7) with no `aria-label`; some screen readers will announce "multiplication sign" instead of a recognizable close action, a minor accessibility regression versus the plain "X" it replaced.

I applied fixes #1, #3, and #4 locally in the working tree (verified via `git stash` diff that this drops ESLint errors on the file from 5 to 2 — the 2 remaining, unused `React` import and unused `message` destructure param, are pre-existing and outside this diff's scope). #2, #5, #6 are left as report-only since they need a product/backend decision or touch code outside this diff.

---
🤖 Automated review

nikshepkulli (approved):
Approved. Pavan correctly fixed the actual bug — the old code fired `notify()`+`closeModal()` regardless of the un-awaited POST's outcome (and swallowed the error), so every submit faked success. Now: `await`, success only on 2xx, failure keeps the modal open with data intact + a WhatsApp fallback, 10s timeout so a dead backend fails visibly, double-submit guard. Traceability clean (`Fixes #2`).

I pushed one commit: the founder confirmed both RCA WhatsApp numbers, so the fallback now offers **both** (80737 62257 + 91104 24403) with a prefilled enquiry message. Build green.

Note: merging this closes #2's *silent-loss* problem; the **backend revive** (so submissions actually land instead of always falling to WhatsApp) continues under https://github.com/Rest-coder-academy/trca/issues/19. Great work @Pavankulli.

Issue #15, opened 2026-09-01
[P1] Batch dates admin-controlled — RCA edits them without a code change


nikshepkulli:
## Why
#3 fixes the immediately stale hardcoded date (24-05-2025). But hardcoding just means it goes stale again in a few months and someone has to redeploy every time. The real fix: **RCA (Uday) updates batch dates themselves, no developer and no code change.**

## Outcome
An authenticated RCA admin can add / edit / remove batch entries, and the **Batches** section on the live site reflects it **without a redeploy**.

## Acceptance criteria
- [ ] A **protected admin screen** (RCA admin only) to manage batch entries — course, start date, mode/timing, seats or status.
- [ ] The Batches section reads from that source, **not** a hardcoded array.
- [ ] Changes show on the live site with no code change / redeploy.
- [ ] **Graceful staleness:** a batch whose date has passed auto-hides (or shows "new dates coming soon") instead of advertising an old date — so it can never look inactive again.

## Depends on
The **backend decision** (revive vs replace `trcabe.onrender.com` — see #2). The storage/mechanism for this should follow whatever we land on there; if the backend is being replaced, build this on the new one. Flag if you're blocked on that call.

## Notes
- Same admin surface can later hold **hero copy** (Abhigna flagged it also needs Uday's sign-off) — out of scope here, but design the admin so it can grow.
- Builds on #3 · part of epic #13.

nikshepkulli:
@Pavankulli kicking this off — it's yours. #3 fixes the stale date now (Abhigna); #15 is the durable layer so RCA edits batch dates themselves without a redeploy. Dependency on the #2 backend decision: we're **reviving `trcabe.onrender.com` now and consolidating into the payment build (PR #1)** — build the admin batch-dates store on whatever backend we land on there, so flag if you're blocked. Ping me for Render access when you need it.

PR #26, opened 2026-09-01
Fix batch dates being hardcoded and advertised as upcoming when stale

### LINKED ISSUE #3: [P0] Batch dates are hardcoded to 24-05-2025

Pavankulli:
## Summary
- The Batches section had every entry hardcoded to `date:"24-05-2025"` — 15 months in the past — advertising the academy as dormant to anyone who read it.
- Add `batchDateUtils.js`: a shared helper (`isBatchUpcoming`, `formatBatchDateShort`, `getNextBatchForCourse`) so date-status logic lives in one place, not scattered across components.
- `BatchesCard`: a batch whose date has passed now shows a muted **"Batch closed"** tag, replaces the date row with "New dates coming soon", and swaps the CTA from "Enquire Now" to **"Join the Waitlist"**. A past date can no longer render as if it's upcoming.
- `CoursesCard`: each course card now shows a **"Next batch · 16 Sep"**-style tag (matched by course name against the batches data), visible without scrolling to the Batches section — or "New dates coming soon" if none are upcoming.
- Updated the batch dates off the stale placeholder to near-future Wednesdays (Sep–Oct 2026), staggered per course. These are still placeholders pending the actual schedule from Uday — flagged with an inline comment in `batches.js` since I don't have the authoritative dates; the closed/waitlist safety net means the site can't misrepresent it again even if these go stale.

Fixes #3

## Test plan
- [x] `npm run lint` shows no new errors (pre-existing repo-wide lint errors unaffected)
- [x] Manually verified in a browser: course cards show the correct "Next batch" date matching the Batches section
- [x] Manually verified in a browser: temporarily backdating a batch entry renders "Batch closed" + "New dates coming soon" + "Join the Waitlist" CTA, then reverted
- [ ] Swap placeholder dates in `batches.js` for Uday's real upcoming schedule before/after merge

### DIFF
--- a/src/components/organism/Batches/Batches.css
+++ b/src/components/organism/Batches/Batches.css
@@ -86,6 +86,18 @@
 {
     margin: 0;
 }
+.batches .batch-status-chip
+{
+    display: inline-block;
+    margin-top: .4rem;
+    padding: .15rem .7rem;
+    border-radius: 999px;
+    background-color: rgba(255, 255, 255, 0.18);
+    color: #ffffff;
+    font-size: .75rem;
+    font-weight: 600;
+    letter-spacing: .02em;
+}
 .batches .card-actions
 {
     border-top: .5px solid rgba(211, 211, 211, 0.582);
--- a/src/components/organism/Batches/BatchesCard.jsx
+++ b/src/components/organism/Batches/BatchesCard.jsx
@@ -9,6 +9,7 @@ import TypoGraphyComponent from '../../atoms/TypoGraphyComponent/TypoGraphyCompo
 import ButtonComponent from '../../atoms/ButtonComponent/ButtonComponent';
 import { List, ListItem, ListItemButton, ListItemIcon, ListItemText } from '@mui/material';
 import BatchItem from './BatchItem';
+import { isBatchUpcoming } from './batchDateUtils';
 import LaptopIcon from '@mui/icons-material/Laptop';
 import CalendarMonthIcon from '@mui/icons-material/CalendarMonth';
 import CalendarTodayIcon from '@mui/icons-material/CalendarToday';
@@ -21,6 +22,7 @@ import { useAuth } from '../../../App';
 
  function BatchesCard({name,date,day,time,trainer,duration,mode,contact}) {
     let {openModal}=useAuth()
+    let upcoming=isBatchUpcoming(date)
   return (
     <Card sx={{}} className="card">
       <Box className="course-header">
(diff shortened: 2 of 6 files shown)


Pavankulli:
@nikshepkulli heads up — the batch dates in this PR (`src/components/organism/Batches/batches.js`) are placeholders I set so the new closed/waitlist logic had realistic upcoming dates to work against. They are not the real schedule. Could you or Uday confirm the actual upcoming batch dates so they can be swapped in before/after merge? Flagging here instead of opening a separate issue since it's a one-line data update once we have the real dates.

Pavankulli:
## Self-review

Ran a self-review pass on this PR. Two real issues came up, both fixed in 9f33d9b:

1. **Stale Day/Time on closed batches** — when a batch's date passed, the Date row correctly switched to "New dates coming soon," but the Day/Time rows below it kept showing the old schedule as if it were still confirmed, contradicting the message. Now the whole date/day/time block collapses to a single "Status" row when closed.
2. **Unvalidated date parsing** — `parseBatchDate` accepted any three numbers in any order with no range check, so a malformed or out-of-order date (e.g. ISO format pasted by mistake) would silently roll over into a nonsense date via JS `Date` overflow instead of being caught. Added range validation and rollover rejection.

Everything else — CSS scoping for the new tags, course/batch name matching, sort/filter logic, timezone handling — checked out clean. Build and lint pass. This PR is set from my end.

nikshepkulli (approved):
Approved — strong work. Not just a date swap: the `batchDateUtils` parsing is careful (DD-MM-YYYY, rollover-guard, fails closed), and the **auto-close safety net** means a stale date can never advertise the academy as active again. Course-card "Next batch" tags + graceful name-mismatch degradation are good touches, and the new dates are consistent Wednesdays. Traceability clean (`Fixes #3`).

One follow-up (not blocking): the dates are placeholders — we need **Uday's real schedule** to swap in. The safety net makes shipping these safe, and merging removes the live 15-month-old date now. #15 (admin-editable dates) is the self-serve layer on top.

Merging — auto-deploy will push it to production.

PR #36, opened 2026-09-01
Fix navbar hamburger being a dead control from 600px up

### LINKED ISSUE #5: [P1] Navbar hamburger is a dead control at 600px and above

Pavankulli:
## Summary

MUI `sx` breakpoints are min-width and cascade upward with no automatic reset. The hamburger's `display:{sm:'block'}` had no override at `md`/`lg`, so it stayed visible at every width from 600px onward — right alongside the inline nav links (also `display:{sm:'block'}`). Two navigations rendered together from 600px up, and the icon was black-on-navy (near-invisible) in that same range.

- Hamburger now visible only below `md` (900px).
- Inline nav links now visible only from `md` (900px) up — matching breakpoint, so exactly one nav shows at any width.
- Drawer's own breakpoint matched to the hamburger's, so the button always opens something.
- Icon color fixed to always white (it sits on the navy AppBar).
- Fixed a latent bug found while in there: `marginLeft:"auto"` was nested *inside* the `display` breakpoint object instead of being a sibling `sx` property, where MUI silently ignores it.
- Drawer rows bumped to a 44px minimum touch target.

Fixes #5

## Test plan
- [x] `npm run lint` shows no new errors (the two Navbar.jsx errors are pre-existing unused imports, untouched by this diff)
- [x] `npm test` — all 42 unit/integration tests pass
- [x] Added `e2e/navbar.spec.js` — verifies exactly one nav (hamburger vs. inline links) is visible at 375/768/1024/1440px, and that the hamburger actually opens the drawer
- [x] Confirmed the new e2e test is a real regression test: ran it against the original code first — it correctly **fails** at 768/1024/1440px (hamburger visible when it shouldn't be), then reran against the fix — all 5 pass
- [x] Full e2e suite (`npm run test:e2e`) — 9/9 pass, no regressions elsewhere


Pavankulli:
## Self-review

Ran a self-review pass on this PR. One real issue came up, fixed in dbc5ee2:

**Hamburger icon was white-on-white.** My original commit hardcoded the icon color to white, reasoning the AppBar background is navy — that assumption was wrong. `Navbar.css` has `nav { background-color: white !important; }`, and since `<AppBar component="nav">` renders as a literal `<nav>` element, that rule wins. The icon was genuinely invisible (verified via computed style: both `rgb(255,255,255)`) at every width this PR now shows it — below 900px — which undercut the whole point of the fix. Corrected to black, matching what the original code already rendered across that same range (its white branch only applied at 1200px+, a width where the hamburger is now hidden entirely).

Added a new e2e test asserting the icon's computed color differs from the navbar's computed background, and confirmed it independently: ran it against the white-icon version first (it correctly failed), then against the fix (passes).

Everything else — breakpoint alignment, the `marginLeft` no-op fix, drawer touch targets — checked out clean. Full suite: 42/42 unit+integration, 10/10 e2e, build and lint pass with no new errors. This PR is set from my end.

Pavankulli:
## Rebased on main

This branch was cut a long time ago (before tokens.css, the trainer/founder/blog/contact/faq/placements pages, the whole SEO epic — 100+ commits behind). Confirmed the bug it fixes is **still live on \`main\`**: the hamburger and inline nav both render from 600px up, exactly as #5 describes.

Merged \`main\` in. Only one real conflict, in the drawer's nav-item click handler:
- This PR's side: react-scroll's \`Link\` directly, plus a \`minHeight: 44\` touch-target fix.
- Main's side: a newer \`goToSection\` handler that's cross-page-aware (added later so nav items work correctly when you're not already on the homepage).

Kept main's \`goToSection\` handler (needed — it's newer and does more) and layered this PR's \`minHeight: 44\` fix on top. Everything else — the \`sm\`→\`md\` breakpoint fix, the icon color fix, the drawer row touch targets — merged in cleanly with no conflict, since main never touched those lines.

## Verification

- \`npm run build\` — succeeds
- \`npm run lint\` — 388 problems on both plain \`origin/main\` and this branch (identical baseline, zero new issues)
- \`npm test\` — 64/64 unit/integration pass
- **Full e2e suite (built app + wrangler pages dev) — 12/12 pass**, including all 6 of this PR's own navbar regression tests (exactly-one-nav-usable at 375/768/1024/1440px, hamburger opens the drawer, icon has contrast against the navbar background) and the pre-existing smoke suite (homepage, batches API, admin auth, checkout CTA, enquiry form) — confirms nothing else broke pulling in this much history.

Should be ready for review now.

Pavankulli:
## Self-review

No correctness bugs found — the fix is logically sound and the new e2e tests exercise the actual regression. Confirmed the old code had no upper bound on the hamburger's visibility (\`{sm:'block', xs:'block'}\` = visible from 600px up forever) while the inline nav also went visible at \`sm\` — the real dead-control overlap. The new \`{xs:'block', md:'none'}\` / \`{xs:'none', md:'block'}\` pair is complementary and matches the e2e assertions.

Three minor, non-blocking observations (nothing fixed, just flagging):

1. The "icon has contrast" e2e test only checks \`iconColor !== navBackground\`, not an actual WCAG ratio — a future change to a near-white gray icon would still pass this test while being unreadable.
2. The \`{xs:.../md:...}\` breakpoint pair is repeated three times (IconButton, inline Box, Drawer) with no shared constant — a future threshold change could easily update only one or two of the three and reintroduce this exact bug.
3. \`Navbar.css\`'s hand-rolled px breakpoints don't align with the MUI \`md=900\` threshold this fix now depends on. Both are currently empty media query bodies so there's no live conflict, but worth knowing if either file gets more rules added later.

None of these block merging — flagging as known follow-ups. Build/lint/64 unit tests/12 e2e tests all still green from the rebase verification above.

Pavankulli:
## Two more fixes found manually re-testing this PR at each breakpoint

**Drawer not closing on nav click (e1d8899):** the drawer's outer \`Box\` already closes on any click inside it via bubbling (\`onClick={handleDrawerToggle}\`, the standard MUI pattern). Each nav item's \`span\` ALSO called \`handleDrawerToggle()\` explicitly before navigating — so a single click fired the toggle twice in the same batched event: closed, then immediately reopened. Net effect: the drawer looked like it never closed after tapping a link. Removed the redundant call so the outer \`Box\`'s bubble-triggered close is the only one firing.

**Logo touching the hamburger at 576-991px (0a6a157):** \`Navbar.css\` already has \`justify-content: space-between\` at the 320-479 and 480-575 breakpoints specifically to push the logo left and the hamburger right when they're the only two visible controls. The next two brackets — 576-767.98 and 768-991.98, exactly the range this PR's \`sm\`→\`md\` fix now shows the hamburger-only layout in — had empty bodies, so nothing pushed them apart there. Extended the same rule to both.

Measured the fix: gap between logo and hamburger now scales properly with viewport width (200px at 375, 453px at 768, 535px at 850), and the hamburger correctly disappears entirely at 900px+.

Re-verified after both fixes: \`npm run build\` succeeds, \`npm run lint\` — 322 problems on both plain \`origin/main\` and this branch (identical baseline, zero new issues; verified via an isolated git worktree this time rather than \`git checkout\`, since that method almost cost me the drawer-close fix mid-session), \`npm test\` 64/64 pass, and the full e2e suite (12/12, including this PR's own navbar regression tests) still green.

nikshepkulli (approved):
Clean, well-scoped fix. The breakpoint logic is correct — hamburger `xs→md:none`, inline links `xs:none→md:block`, matched at 900px so exactly one nav renders at any width — and it's even better-justified now that the nav carries the extra For Parents / About links from #117 (7 inline items need the hamburger up to a wider breakpoint). Nice catches on the nested `marginLeft` (MUI silently ignored it) and the black-on-navy invisible icon. The `e2e/navbar.spec.js` is a real regression test (verified failing on the old code first). Compatible with the #117 navbar rewrite (GitHub confirms CLEAN). CI green. `Closes #5` verified.

Minor: the PR summary says "icon fixed to always white" but the code sets it black (`#000`) — the code is right (navbar bg is white); just a stale line in the description. Merging.

Issue #39, opened 2026-09-01
[Feature] FDE course + Enroll Now funnel + Razorpay (EMI) checkout


nikshepkulli:
## What

Introduce **FDE (Forward Deployed Engineering)** — a ₹50,000 course taught by Nikshep — and turn the site from enquiry-first to **enrolment-first**.

- New **FDE** course card (flagship, priced, syllabus), taught by the founder.
- Every course CTA: **Enquire Now → Enroll Now**, funnelling into enrolment (course cards + batch cards).
- **FDE → Razorpay checkout** with **EMI** (bank-EMI: student's bank splits it, academy receives ₹50k upfront). Other courses → free 'register interest'.
- **/admin/enrollments** to see paid + registered enrolments.

## Money rules (from design PR #1, enforced in code)
- Amount decided **server-side** (`shared/enroll.js`), never from the browser.
- Razorpay **signature verified server-side**; recording is **idempotent** (one row per order).
- Referral (`?ref=`) rides into Razorpay `notes` + the row.
- **Inert until keys**: with no RAZORPAY_KEY_ID/SECRET, checkout falls back to recording interest — nothing breaks.

## Done when
FDE is live with a working Enroll → pay flow (once Razorpay keys + EMI are enabled and the `enrollments` table is created), and other courses funnel to register.

Issue #10, opened 2026-09-01
[P2] Placement company logos are hotlinked from third-party servers


Abhigna1975:
`src/components/organism/placements/placement.js` sets each student's company logo to a URL on someone else's server:

- `https://community.sap.com/legacyfs/online/storage/blog_attachments/2016/07/sap_hybris_blue_800_995491.jpg`
- `https://www.equitybulls.com/equitybullsadmin/uploads/HCL%20Technologies%20Limited.jpg`

These are hotlinks to a 2016 SAP blog attachment and an unrelated financial news site. They can 404, get moved, or be swapped for something else without warning, and the placements section — the most persuasive thing on the site — would render broken images.

### What to build

- [ ] Download each company logo, compress it, and commit it to `src/assets/clients/` (several are already there).
- [ ] Point `placement.js` at the local imports.
- [ ] No remote image URL anywhere in `src/`.
- [ ] Fallback: if a logo is missing, render the company name as text rather than a broken image.

### While in here

`src/assets/clients/` has both `.jpg` and `.png` of the same logo for Accenture, Cognizant, Capgemini, Tech Mahindra and Wipro. Keep one each.

### Also worth raising with Nik

The section is titled **Clients** and shows Wipro, Infosys, Accenture, HCL, Capgemini, Cognizant, Mindtree and Tech Mahindra. These are companies students were *placed at*, not clients of the academy. "Our students work at" is both accurate and more persuasive. Not a UI decision — flagging it, not changing it.

### Done when

No third-party image URL is loaded anywhere on the site.

PR #49, opened 2026-09-01
fix(button): all six states — hover, pressed, focus, disabled, loading (#7)

### LINKED ISSUE #7: [P1] ButtonComponent has no hover, pressed, focus or loading state
`src/components/atoms/ButtonComponent/ButtonComponent.jsx` is the button used everywhere — Register Now, Apply Now, Enquire Now, the form submit. It defines exactly one state: a disabled colour of `#706c61`. No hover, no pressed, no focus, no loading.

The practical results: nothing responds to the cursor, keyboard users get no visible focus, and the submit button can be pressed repeatedly during a request with no feedback — which is half of why #2 goes unnoticed.

### What to build

All six states from section 3 of `design/ui-template.html`, each a distinct fill — no opacity tricks:

- [ ] default `--rca-navy` `#03084C`
- [ ] hover `--rca-navy-hover` `#0A1270`
- [ ] pressed `--rca-navy-pressed` `#01062F`
- [ ] focus — 3px `rgba(20,99,137,.40)` ring, visible on keyboard focus
- [ ] disabled `--rca-disabled-bg` with `--rca-disabled-ink` label
- [ ] loading — spinner, label swaps to `Sending…`, **width does not change**, button disabled

Plus:
- [ ] Height `48px` desktop / `44px` mobile, from `--rca-control-h`.
- [ ] `border-radius: var(--rca-radius-md)`, replacing the `4px` default and the `borderRadius='0px'` passed at each call site.
- [ ] Transition on background only, `180ms cubic-bezier(0.2,0,0,1)`. No lift, no scale.
- [ ] Same treatment for the outline variant.
- [ ] Drop `disableFocusRipple` — it's removing the only focus affordance the button had.

### Done when

Every button on the site shows all six states, and the site is navigable and operable by keyboard alone.

Abhigna1975:
Closes #7

`ButtonComponent` is the button used everywhere — Register Now, Apply Now, Enquire Now, the enquiry submit. It defined exactly one state (a disabled fill of `#706c61`). No hover, no pressed, no focus, no loading.

## The six states

All from `design/tokens.css`, each a distinct fill — no opacity tricks. Verified in a real browser rather than by reading the CSS:

| State | Spec | Measured |
|---|---|---|
| default | `#03084C` | `rgb(3, 8, 76)` ✅ |
| hover | `#0A1270` | `rgb(10, 18, 112)` ✅ |
| pressed | `#01062F` | `rgb(1, 6, 47)` ✅ |
| focus | 3px `rgba(20,99,137,.40)` | `rgba(20,99,137,0.4) 0 0 0 3px` ✅ |
| disabled | `--rca-disabled-bg` / `--rca-disabled-ink` | ✅ |
| loading | spinner, `Sending…`, width unchanged | ✅ |

Plus: height **48px** desktop / **44px** mobile from `--rca-control-h`; radius `--rca-radius-md` (8px), replacing the `4px` default and the `borderRadius='0px'` passed at each call site; transition on `background-color`/`border-color` only at 180ms — **no transform**, no lift, no scale; the outline variant gets the same treatment; `disableFocusRipple` is gone, since it was removing the only focus affordance the button had.

## What the latest commit fixes

**The enquiry form never used the `loading` prop it was built for.** It hand-rolled a spinner and passed `disabled={isSubmitting}`, so the submit button went *grey* mid-request — the disabled fill, the exact thing the ticket rules out. It now passes `loading`.

Measured across a stalled request:

```
before   w=141.14  bg=rgb(3, 8, 76)   "Submit"    disabled=false
loading  w=141.14  bg=rgb(1, 6, 47)   "Sending…"  disabled=true  aria-busy=true
WIDTH STABLE: true
POSTs: after 1st click 1, after 2nd click 1  -> double-submit blocked
```

Both labels are always rendered, stacked in one grid cell with one hidden, so the button is always as wide as the longer of the two and going into loading cannot make the layout jump.

**A keyboard focus floor.** The done-criterion is that the site is operable by keyboard alone, and it was not. react-slick's carousel arrows and dots are real `<button>`s in the tab order that ship with `outline: none`, so **eleven controls on the homepage showed nothing when focused**. A zero-specificity `:where(...):focus-visible` rule in `tokens.css` puts a ring under everything without overriding a control that draws its own — the button keeps its own 8px ring (verified: `radius=8px shadow=rgba(20,99,137,0.4) 0 0 0 3px`).

Tab sweep of the homepage, before and after:

```
before: NO VISIBLE FOCUS on Previous, Next, dots 1-9
after:  every tab stop has a visible focus indicator (60 stops)
```

Restyling those arrows properly is #106, and this does not pre-empt it — it only guarantees they are never invisible.

**Drops a dead `textColor` prop** left at the navbar call site after the component stopped accepting it.

## Checks

- `npm run test` — 75 passed, 7 files
- `npm run build` — clean
- `npm run lint` — no new errors on tracked files (the repo has 308 pre-existing on `main`)
- `main` merged in; this branch was 9 commits behind

## Noted, not fixed here

`android/` from the Capacitor work in #109 is **neither committed nor gitignored**, so ESLint walks its minified build output and reports ~680 extra errors locally. Belongs to #109 / PR #115, not this ticket.

https://claude.ai/code/session_01DjmEGczffCE2uiKAtdiXuE


nikshepkulli:
Landed the foundation: **#14 (design tokens/template) and #52 (adopt tokens — one typeface, four radii) are merged to `main`**. This PR now **conflicts with #52** — it was branched before the token refactor, so it touches styles that #52 rewrote.

Could you **rebase onto `main` and re-apply your design changes on top of the new tokens**? I did not auto-resolve because it needs your design intent (a naive resolution could quietly undo the token adoption). CI (test + e2e + build) will confirm once rebased. Also merged: #51 (#10 placement logos), and kulli&co #185 (#39 headshots). Thanks!

Pavankulli:
## Rebased on main

This branch was cut a long time ago — before the entire premium-card redesign, the SEO epic, trainer credibility cards, everything. Confirmed the bug it fixes is **still live on \`main\`**: \`ButtonComponent\` still uses the old \`makeStyles\`-based single disabled state, still has \`disableFocusRipple\` removing the only focus affordance, still takes a hardcoded \`borderRadius="4px"\` default.

Merged \`main\` in. Five conflicts, each resolved by figuring out which side was actually current:

- **\`design/tokens.css\`** (add/add — both branches independently created this file): took main's version, a strict superset (adds shadow-modal, translucency, and WhatsApp brand tokens this branch's copy predates).
- **\`ButtonComponent.jsx\`**: the conflict was just the seam between this PR's full rewrite and main's still-unfixed original — everything below it had already auto-merged into the new six-state implementation. Kept this PR's doc comment, discarded main's leftover \`useStyles\`/\`makeStyles\` block.
- **\`Banner.css\`**: main still has a \`!important\` hover override that fights the button's own hover state — same bug this PR already targets, just with the color value changed since (\`var(--rca-blue)\` instead of the raw \`rgb(31,31,211)\` the original comment names). Kept the removal, updated the comment to not claim a stale color value.
- **\`CoursesCard.jsx\`**: completely rewritten on main since this branch was cut — the premium-card redesign replaced \`ButtonComponent\` entirely with raw \`<button>\` elements (\`cc-book\`/\`cc-counsellor\`). This PR's changes to that file are fully superseded; took main's version wholesale.
- **\`main.jsx\`**: this branch predates \`BrowserRouter\`/\`ThemeProvider\`/self-hosted fonts entirely. Took main's side (all of it is required — used further down in the same file).

## Verification

- \`npm run build\` — succeeds
- \`npm run lint\` — 320 problems on this branch vs 322 on plain \`origin/main\` (verified via an isolated git worktree) — **2 fewer**, consistent with removing the dead \`makeStyles\`/\`DeleteIcon\` imports the old implementation carried; not a red flag.
- \`npm test\` — 64/64 pass
- Manually verified all remaining call sites (\`Navbar.jsx\`, \`BannerContent.jsx\`, \`EnquiryForm.jsx\`, \`BatchesCard.jsx\`) had their now-redundant \`borderRadius=\"0\"\`/\`bgColor=\"bg-btn-blue\"\` props removed, matching this PR's original scope exactly.
- Ran a headless Playwright check against the real hero "Register Now" button: default \`rgb(3,8,76)\` (navy), hover \`rgb(10,18,112)\` (navy-hover, confirmed changing on :hover), focus-visible ring \`rgba(20,99,137,0.4) 0 0 0 3px\`, border-radius \`8px\` (token-based, not the old hardcoded value), height \`48px\` — all match the token values this PR describes.

Should be ready for review now.

Pavankulli:
## Self-review

Two real regressions found and fixed in eea9c90:

**Fixed — nav link text color regression:** the \`ButtonComponent\` rewrite dropped support for the old \`textColor\`/\`className\` props entirely, so \`Navbar.jsx\`'s nav links (the only \`variant=\"text\"\` usage anywhere in the app) silently fell through to MUI's default primary blue instead of the intended dark ink — a visible, site-wide color regression. Gave \`VARIANT_SX.text\` a real \`color: var(--rca-ink)\` and a subtle hover tint, removed the now-dead \`textColor=\"color-dark-black\"\` prop from the call site, and corrected the component's own comment, which still claimed the navbar sits on a navy bar — it's white, per \`Navbar.css\`'s own \`background-color\` rule.

**Fixed — stale \`!important\` hover override:** \`Navbar.css\` still had two legacy rules (\`nav button:hover\`, \`nav button:last-child:hover\`) forcing "Apply Now" to \`--rca-blue\` on hover, defeating the \`--rca-navy-hover\` this PR gives every contained button. Exact same class of bug this PR already fixed once in \`Banner.css\` (an \`!important\` override beating the button's own declared state) — just a second instance of it, in the file right next door. Removed both rules.

**Noted, not fixed** — four lower-severity observations, none blocking:
- The \`loading\`/\`loadingLabel\` props this PR adds have no caller yet (\`EnquiryForm.jsx\`'s real submit button still uses its own \`disabled\`+\`CircularProgress\` pattern). This is intentional and already flagged in the PR's own description: *"The loading prop has no consumer yet. Wiring it to the enquiry submit... is #2, and it's where this belongs."* Not a regression — the enquiry form's existing double-submit protection still works, just via a different mechanism.
- \`VARIANT_SX.outlined\` hardcodes \`rgba(3, 8, 76, 0.06)/0.12\` as magic literals instead of a token. No matching tint token exists in \`tokens.css\` yet — adding one is a design-system decision beyond this component fix.
- The custom \`SPINNER_SX\` keyframe spinner duplicates what MUI's \`CircularProgress\` already provides elsewhere (\`EnrollForm.jsx\`, \`EnquiryForm.jsx\`). Cosmetic/consistency point, not a bug, given \`loading\` has no live caller yet anyway.
- Every button now renders a few extra wrapper \`Box\` nodes to support layout-stable loading, even though nothing currently triggers loading. Minor, permanent DOM overhead for a currently-inactive feature.

Re-verified after the two fixes: \`npm run build\` succeeds, \`npm run lint\` unchanged at 320 (0 new issues), \`npm test\` 64/64 pass. Manually confirmed in the browser: nav link "Courses" now renders \`rgb(51,51,51)\` (\`--rca-ink\`, not MUI's \`rgb(25,118,210)\`), and "Apply Now" now hovers to \`rgb(10,18,112)\` (\`--rca-navy-hover\`, not the old \`rgb(20,99,137)\` \`--rca-blue\`).

nikshepkulli:
@Abhigna1975 now **CONFLICTING** after recent merges — please rebase onto latest main. The ButtonComponent states (#7) are still needed. Small PR, should be a quick rebase.

Issue #11, opened 2026-09-01
[P2] Responsive pass at 375 / 768 / 1024 / 1440


Abhigna1975:
Every CSS file has a `@media` block, but the breakpoints were added per component as problems appeared rather than to a plan, and the widths don't agree with each other. The reviews carousel alone uses `1024`, `600` and `479.98`, while MUI's `sx` props elsewhere use `xs`/`sm`/`lg`. #5 is one symptom; there are likely more.

Most traffic to a Bangalore coaching site arrives on a phone, so 375 is the primary target rather than an afterthought.

### What to build

Walk the whole site at the four widths in section 7 of `design/ui-template.html` and fix what breaks:

- [ ] **375** — 1 column, drawer nav, 44px controls, 16px gutter, H1 32px
- [ ] **768** — 2 columns, drawer nav, 24px gutter
- [ ] **1024** — 3 columns, inline nav, hamburger hidden
- [ ] **1440** — container capped at 1200px, centred

Checks at each width:
- [ ] No horizontal scroll — compare `document.body.scrollWidth` to `clientWidth`; if it's wider, find the element whose right edge is past the viewport.
- [ ] No text under 16px in body copy, nothing clipped, no overlaps.
- [ ] Every tap target 44px or more, floating WhatsApp button included.
- [ ] Section padding on the spacing scale, not per-component one-offs.
- [ ] `Courses.jsx` uses `mx={12}` — 96px of fixed horizontal margin at every width, including a 375px phone. Replace with the responsive gutter.

### Done when

All four widths are clean, screenshots attached to the PR.

PR #52, opened 2026-09-01
refactor(styles): adopt the design tokens — one typeface, four radii (#6)

### LINKED ISSUE #6: [P1] Adopt design tokens: one typeface, four radii

Abhigna1975:
Closes #6. **Stacked on #14** — base is `design/ui-foundation`; GitHub retargets to `main` once #14 merges.

## The done-when is met

```
$ grep -rE "#[0-9a-fA-F]{6}|rgb\(" src/
(nothing outside src/assets/new logo.svg)
```

It was **164 colour literals across 25 files, 72 of them distinct**. The only ones left are the `fill` attributes inside the logo SVG, where the colour belongs to the asset — it's loaded as an ``, so a CSS variable wouldn't resolve there.

## One typeface — three things were stopping it

The ticket names two. There was a third, and it's why fixing the first two wasn't enough:

1. `index.css` forced `'Segoe UI'` with `!important` on **every element**.
2. The navbar and footer set **Tilt Neon**, also `!important`.
3. **MUI's default theme sets Roboto** on every `Typography`, `Button` and input, at a specificity no `*` rule can reach. With the CSS fixed, the computed body font was still `Roboto`. `src/theme.js` sets the family from the token; `main.jsx` wraps the app in the provider.

### And Montserrat was never actually being downloaded

`src/styles/fonts.css` carries the `@import` — but **nothing imports that file**. Same for `global.css` and `width.css`: all three are dead. It rendered on my machine only because Montserrat happens to be installed locally; the recorded font requests showed Dancing Script and Tilt Neon going out and no Montserrat.

`index.html` now requests it directly. Worth a follow-up ticket that `.ff-MS`, the `.fs-*` / `.fw-*` utilities and the Toastify brand overrides in `global.css` have never applied to anything.

**Dancing Script** is gone — loaded on every page load, used nowhere.

## Radii: ten → four

`.3rem`, `3px`, `5px`, `.4rem`, `.5rem`, `.6rem`, `10px`, `12px`, `30px`, `0.8rem` map onto `sm` / `md` / `lg` / `pill`. `50%` and `100%` stay: those are avatar circles, not a radius scale.

## Gutters

All seven sections used `mx={12}` — **96px of fixed horizontal margin at every width**, which on a 375px phone left barely half the screen for content. Replaced with `.rca-section`: the container and gutter tokens, capped at 1200px and centred. (This also settles the `Courses.jsx` item on #11.)

**That change surfaced a bug worth naming.** Capping the container gave the page a horizontal scrollbar at 768 and up. The slick carousels render a track wider than their container and had been relying on those 96px margins to absorb the overhang — including a next-arrow sitting 50px past the viewport. `.rca-section` clips it. Baseline `main` measured clean, so this was mine to fix, not pre-existing.

## Five tokens added

For the cases that genuinely need alpha and had none: `--rca-shadow-modal`, `--rca-tint-on-dark`, `--rca-tint-on-dark-strong`, `--rca-scrim`. Plus `--rca-whatsapp`, named so it's obvious why a third-party brand green sits outside the scale.

`colors.css` is now a thin layer over the tokens rather than a second source of colour.

## Test plan

Measured at all four widths, against a `main` baseline captured the same way:

| width | h-scroll | body font |
|---|---|---|
| 375 | none ✓ | Montserrat |
| 768 | none ✓ | Montserrat |
| 1024 | none ✓ | Montserrat |
| 1440 | none ✓ | Montserrat |

*(`main` baseline: no h-scroll, body font `"Segoe UI"`.)*

- [x] `npm run build`
- [x] `npm test` — 61 passed
- [x] `npm run lint` — no new problems
- [x] Montserrat confirmed **requested over the network**, not just resolved locally
- [x] Screenshots at 375 / 768 / 1024 / 1440

https://claude.ai/code/session_019LRjAtspWxKEvsjXG6S2gS

### DIFF
--- a/design/tokens.css
+++ b/design/tokens.css
@@ -79,6 +79,19 @@
   /* --- Elevation -------------------------------------------------------- */
   --rca-shadow-card:  0 1px 3px rgba(3, 8, 76, 0.08);
   --rca-shadow-hover: 0 6px 18px rgba(3, 8, 76, 0.10);
+  --rca-shadow-modal: 0 20px 60px rgba(3, 8, 76, 0.25);
+
+  /* --- Translucency ------------------------------------------------------
+     The only places a colour legitimately needs alpha: tints over the navy
+     cards, and the dialog scrim. Everything else is a solid token.          */
+  --rca-tint-on-dark:        rgba(255, 255, 255, 0.15);
+  --rca-tint-on-dark-strong: rgba(255, 255, 255, 0.24);
+  --rca-scrim:               rgba(3, 8, 76, 0.55);
+
+  /* --- Third-party brand -------------------------------------------------
+     WhatsApp's own green. Not ours to restyle, and not part of the palette —
+     named so it is obvious why it sits outside the scale.                   */
+  --rca-whatsapp: #128C7E;
 
   /* --- Controls ---------------------------------------------------------- */
   --rca-control-h:        48px;  /* button / input height, desktop */
--- a/index.html
+++ b/index.html
@@ -22,12 +22,8 @@
 />
     <!-- One typeface (#6). Dancing Script was loaded here and used nowhere,
          Tilt Neon was the navbar and footer face; both are gone. Montserrat is
-         requested here rather than from src/styles/fonts.css, because nothing
-         imports that file — see the PR. Without this link the site falls back
-         to whatever the visitor happens to have installed. -->
-    <link rel="preconnect" href="https://fonts.googleapis.com">
-    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
-    <link href="https://fonts.googleapis.com/css2?family=Montserrat:wght@300;400;500;600;700&display=swap" rel="stylesheet">
+         self-hosted via @fontsource/montserrat, imported in src/main.jsx — no
+         font request leaves the site. -->
 <!-- <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.3/dist/css/bootstrap.min.css" rel="stylesheet" integrity="sha384-QWTKZyjpPEjISv5WaRU9OFeRpok6YctnYmDr5pNlyT2bRjXh0JMhjY6hW+ALEwIH" crossorigin="anonymous"> -->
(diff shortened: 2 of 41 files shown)


Abhigna1975:
Pushed: **Montserrat is now self-hosted** (`@fontsource/montserrat`, five weights) instead of a `fonts.googleapis.com` link.

No font request leaves the origin any more — verified by recording every non-local request on a full page load, and by `document.fonts` reporting Montserrat 300/400/500/600/700 all loaded.

Since `styles/fonts.css` isn't imported by anything, its stale Google `@import` was the thing that made the face *look* loaded when it wasn't. The file now says so at the top rather than leaving the next person to find out the same way.

Issue #42, opened 2026-09-01
[Portal] Phase 1 — Student SSO (Google + Microsoft) + Capacitor app shell + student home


nikshepkulli:
Part of #41.

The foundation + the "turn it into an app quickly" proof: a real installable Android app you log into with Google or Microsoft.

**Auth — all-Cloudflare (no Supabase):** hand-rolled **OIDC Authorization Code + PKCE** in Pages Functions.
- `GET /auth/:provider/start` → redirect to Google / Microsoft consent (state + PKCE).
- `GET /auth/:provider/callback` → exchange code for tokens server-side, **verify the ID token against the provider JWKS**, upsert the user in **D1** (`users` table, role default `student`), issue a **signed HttpOnly session cookie** (JWT via Web Crypto, `SESSION_SECRET`).
- `GET /auth/me` / `POST /auth/logout`.
- Login screen (Google + Microsoft buttons) reusing the app brand/atoms; protected `/portal` routes; **student home** shell.
- **Capacitor** added → Android app builds, opens to login; OAuth works in the native shell via deep-link redirect.
- **Inert-until-configured**: without the OAuth client secrets the portal shows "coming soon"; the marketing site is untouched.

**Owner-provisioned (I cannot create OAuth apps or handle secrets):**
- **Google OAuth client** (Google Cloud Console) → `GOOGLE_CLIENT_ID` / `GOOGLE_CLIENT_SECRET`.
- **Microsoft app registration** (Entra) → `MS_CLIENT_ID` / `MS_CLIENT_SECRET` / `MS_TENANT` (common).
- Redirect URIs (web + Capacitor deep link) + a `SESSION_SECRET` Pages secret.

Done when: install the app, tap Sign in with Google/Microsoft, land on the student home.

nikshepkulli:
@Abhigna1975 — delegating Phase 1 (and the epic #41) to you. Here's a clean hand-off.

**Approach (locked):** reuse the existing React app, add Google + Microsoft SSO, wrap with **Capacitor** (the Starteryou playbook). Backend **all-Cloudflare** — D1 + R2 + hand-rolled OIDC (no Supabase). Marketing site + enrolment stay untouched.

**Head-start already on `feat/portal-phase1-auth`** (pushed): session backbone is built + tested — `schema-users.sql` (users table) + `shared/auth.js` (HS256 signed-cookie sessions via Web Crypto) + 8 passing tests. Build the OIDC flows on top of this.

**Phase 1 to build:**
[...]

Issue #113, opened 2026-09-02
[Portal] Google + Microsoft OIDC endpoints (Authorization Code + PKCE)


Abhigna1975:
Part of #42 (epic #41).

Hand-rolled OIDC in **Pages Functions** — no auth SaaS.

- `GET /auth/:provider/start` → build the consent redirect with **state + PKCE** (`code_challenge` S256); stash the verifier in a short-lived HttpOnly cookie.
- `GET /auth/:provider/callback` → validate `state`, exchange the code **server-side**, and **verify the ID token against the provider JWKS** (issuer, audience, expiry, nonce). Never trust an unverified token.
- Upsert the user in D1 (see the schema ticket), then hand off to session issuance.
- Providers: **Google** (`GOOGLE_CLIENT_ID` / `GOOGLE_CLIENT_SECRET`) and **Microsoft Entra** (`MS_CLIENT_ID` / `MS_CLIENT_SECRET` / `MS_TENANT=common`).
- Redirect URIs for both the **web** origin and the **Capacitor deep link**.
- **Inert-until-configured:** with the secrets absent the endpoints return a clean "not configured", never a 500.

**Owner-provisioned (not in this ticket):** the Google Cloud OAuth client and the Entra app registration + their secrets.

**Done when:** a real Google and a real Microsoft account each complete the round trip and land a verified user row in D1.

Issue #112, opened 2026-09-02
[Portal] D1 schema for the portal — users, roles, sessions


Abhigna1975:
Part of #42 (epic #41).

The data foundation for portal auth. All-Cloudflare — **D1**, no Supabase.

- New migration `schema-portal.sql` alongside the existing `schema-*.sql` files.
- `users`: id, provider (`google` | `microsoft`), provider_subject (unique per provider), email, name, avatar_url, **role default `student`**, created_at, last_login_at.
- `sessions` (or a documented reason to stay stateless with a signed cookie only): id, user_id, issued_at, expires_at, revoked_at.
- Indexes on `(provider, provider_subject)` and `email`.
- Upsert helper used by the callback: same provider subject logs in as the same user, never a duplicate row.
- Wire the D1 binding in `wrangler.toml`; document applying the migration locally and remote.

**Done when:** the migration applies cleanly on a fresh D1 and a repeat login upserts rather than duplicates.

nikshepkulli:
@Pavankulli — heads up before you finalise this schema:

Filed **#146** which adds a `parent` role and a `parent_students` link table. This needs to land as part of your #112 schema PR, not as a follow-up migration — retrofitting a role after auth is live is significantly more painful than adding it as one more line in the CHECK constraint now.

The changes to your existing spec:
- `users.role` CHECK constraint accepts `parent` (in addition to student / instructor / admin)
- New `parent_students(id, parent_user_id, student_user_id, relation, created_at)` table with a unique index on the pair
- No seed data — admin creates parent-student links manually via the admin UI

Full spec in #146. When you open the PR, please reference both #112 and #146 in the description so we can close them together.

@Abhigna1975 — you'll need to add one route stub in your PR #127 flow: after OIDC callback, if `role === 'parent'`, redirect to `/portal/parent` (which for now just renders a placeholder "Your child's dashboard is coming soon" — the actual content is #147, ships later). Everything else in #127 stays as-is.

Issue #110, opened 2026-09-02
[Portal] Login screen + protected /portal routes


Abhigna1975:
Part of #42 (epic #41).

The student-facing entry to the portal, reusing the existing brand atoms — no new design language.

- `/portal/login` with **Sign in with Google** and **Sign in with Microsoft** buttons, hitting `GET /auth/:provider/start`.
- Route guard: `/portal/*` redirects to the login screen when there is no session; calls `GET /auth/me` on load.
- Loading + error states (network failure, cancelled consent, unknown user).
- **Inert-until-configured:** with no OAuth client ids the screen shows "coming soon" instead of dead buttons.
- Mobile-first at 375px; large touch targets; works inside the Capacitor shell.

**Done when:** tapping a provider button leaves for consent and, on return, lands inside `/portal`.

PR #127, opened 2026-09-03
docs(portal): OAuth provisioning runbook for the owner (#42)

Abhigna1975:
The portal UI this PR originally carried has since shipped on `main` via #167. After merging `main` in, what remains here is one file.

## What this adds

`docs/OAUTH-SETUP.md` — the console walkthrough behind the README's secrets table, written for whoever holds the academy's Google and Microsoft accounts. Creating OAuth clients is an owner job that cannot be done from the repo, and it is the long-lead item on #42: provider review and tenant permissions are wall-clock waits unrelated to our code.

It covers the Google consent screen (External, and **only** `openid`/`email`/`profile` — any extra scope drops the app into a manual verification queue that takes weeks and buys nothing), the Entra registration (`common`, so students with personal Outlook accounts are not locked out), the six Pages secrets, and how to verify.

Plus the two things that will bite: the Microsoft client secret expires and Google's does not, and an "Android" OAuth client type is the wrong choice — a custom scheme cannot be a registered redirect for a confidential client, so both providers redirect to this origin and the native shell is handed back via the `rca://` deep link at the end of the callback.

## One correction against the landed backend

The doc claimed `curl /auth/me` returns 503 when unconfigured. It does not — `me.js` answers **401** either way, and the 503 lives on `/auth/:provider/start`. The verification section now reads the `providers` array in the 401 body instead, and notes that `isConfigured()` also requires `SESSION_SECRET`, so that array stays empty when the client ids are set but the session secret is not.

## What was dropped in the merge

- `session-contract.js` + test — **deleted, not merged.** They decode a flat `/auth/me` body (`id`, `avatar_url`); the landed endpoint returns `{authenticated, providers, user:{..., picture}}`. Keeping it would mean tests passing against a shape nothing returns.
- The course APIs at this branch's tip — byte-identical to what is already on `main`.
- All eight conflicted files — resolved to `main`'s version verbatim.

## Verification

- 177/177 tests pass; `npm run build` clean.
- Merged tree differs from `main` by exactly `docs/OAUTH-SETUP.md`.

Part of #42. The code side of #42 is now entirely on `main`, Capacitor shell included — what is left is an owner working through this doc.


nikshepkulli (changes requested):
**Overall:** Genuinely careful work. PKCE, CSRF state, JWKS-verified ID tokens with `alg`+`kid` gates, iss/aud/exp/nbf claims validation, HttpOnly+Secure+SameSite cookies, POST-only logout, `no-store` on `/auth/me`, offline-vs-signed-out distinction, and 23 tests that lock in the important properties (session forgery, expiry, tamper, open-redirect guard, PKCE determinism, 503-when-inert). The prose in the PR body reads like someone who has actually debugged an OAuth flow before.

One blocker, plus a handful of defence-in-depth items I'd take as follow-ups.

---

### 🔴 Blocker — backslash bypasses the `next` open-redirect guard

The filter in both `start.js:34` and `callback.js:93`:

```js
next.startsWith("/") && !next.startsWith("//") ? next : "/portal"
```

catches `//evil` and `https://evil`, but **not** `/\evil`. Empirically (Node's URL parser, which matches browser WHATWG behaviour):

```
"/\evil.example"   filter passes → next="/\evil.example"
new URL("/\evil.example", "https://restcoderacademy.in").href
  → "https://evil.example/"       ← open redirect
```

`\` is treated as `/` inside the URL authority segment for HTTP-scheme URLs. Attack:

`restcoderacademy.in/auth/google/start?next=/\evil.example` → user completes a legit Google login → 302 to `/\evil.example` → browser resolves to `https://evil.example/`. Phishing gold, because the initial URL is on a trusted domain and the redirect follows a real signed-in flow.

Same bypass fires in `callback.js` because the cookie carries the value `start.js` already accepted.

**Fix (either file's validator; ideally both, defence in depth):**

```js
// Starts with / but the very next character isn't / or \
const next = /^\/(?![/\\])/.test(requested) ? requested : "/portal";
```

Or with the URL parser (also handles control chars, tab-in-URL, etc.):

```js
let next = \"/portal\";
try {
  const parsed = new URL(requested, \"https://sentinel.invalid\");
  if (parsed.origin === \"https://sentinel.invalid\") {
    next = parsed.pathname + parsed.search + parsed.hash;
  }
} catch { /* fall through */ }
```

Please add one test that both branches (`start.js` cookie value **and** `callback.js` `safeNext`) reject `/\evil.example`.

---

### 🟡 Recommend (follow-ups, not blocking this one)

**1. Add `nonce` to the OIDC flow.**  For code+PKCE with server-side exchange over TLS, replay is already blocked by `code_verifier`. Nonce adds one more mint-and-check that closes the token-substitution edge case (ID token leaked from logs / a TLS-terminating proxy / a provider bug getting injected here). Store it next to `rca_oauth_state`, echo through as `?nonce=` in `start.js`, verify `claims.nonce === expected` in `verifyIdToken`.

**2. When `MS_TENANT` is pinned, enforce the issuer contains that tenant.**  `shared/oidc.js:177` currently accepts any issuer under `login.microsoftonline.com/`, which is right for `common` but wrong once someone sets `MS_TENANT=contoso` — a token from a different Entra tenant with a matching client_id would pass. Something like:

```js
if (!cfg.issuers) {
  const expected = env.MS_TENANT && env.MS_TENANT !== \"common\"
    ? new RegExp(`^https://login\\.microsoftonline\\.com/${env.MS_TENANT}/`)
    : /^https:\\/\\/login\\.microsoftonline\\.com\\//;
  if (!expected.test(String(p.iss))) return null;
}
```

(Needs plumbing `env` into `providerConfig` — small refactor.)

**3. Validate the session JWT header's `alg` in `verifySession`.**  Decorative today — you're the only signer, HMAC-SHA256 output is length-fixed, a bogus signature won't verify. But if a code path ever accepts asymmetrically-signed tokens on the same helper, alg-confusion becomes reachable. Two lines:

```js
const hdr = JSON.parse(dec.decode(b64urlToBytes(header)));
if (hdr.alg !== \"HS256\") return null;
```

**4. Timeout + cache the JWKS fetch.**  `verifyIdToken` fetches JWKS without an abort signal, so if Google stalls the callback hangs until CF's wall-clock cap. `AbortSignal.timeout(5000)`. And a module-level cache respecting the response's `Cache-Control` (both providers set ~24h) removes one round-trip per login on the hot path.

---

### ⚪ Small (not requesting changes)

- `me.js:29` — `session.role || \"student\"` shadow-defaults on the way out. Schema already `NOT NULL DEFAULT 'student'`; the fallback only helps for a JWT minted before that column existed. Fine either way, but a one-line comment saying so would document the intent.
- No `SESSION_SECRET` rotation story. Rotating it invalidates every session instantly — doc-only for now; a `v` claim + secret-list verify is a Phase-2 concern.
- `role` in the JWT means role changes in D1 don't take effect until re-login. `schema-users.sql:26–29` already flags this trade-off for when the portal starts touching payments/admin surfaces — good.

---

### Verdict

Approve after the backslash fix + one test. The other four are legitimate improvements I'd take in follow-up tickets rather than block this on — the PR is otherwise ready and Phase 2 depends on it.

nikshepkulli:
@Abhigna1975 — small heads-up while this PR is still open:

Filed **#146** for parent role in the schema (Pavan is folding it into his #112 work). To keep #127 aligned so we don't need a rework, please add one branch in your post-OIDC route-decision code:

\`\`\`
switch (session.role) {
  case 'student':     return redirect('/portal/home');   // your existing path
  case 'admin':       return redirect('/portal/admin');  // if you have one
  case 'parent':      return redirect('/portal/parent'); // NEW
  case 'instructor':  return redirect('/portal/teach');  // deferred, fine to placeholder
  default:            return redirect('/portal/home');
}
\`\`\`

And a placeholder page at `/portal/parent` that just renders "Your child's dashboard is coming soon — check back in a few weeks." The actual dashboard content is scoped in #147, which is Q4 work.

Everything else here looks good on my initial read — will do a full review pass separately.

Abhigna1975:
The backslash bypass is fixed in `54358a4`. I confirmed it first rather than taking it on faith — `/\evil.example` passed the old filter, and `new URL("/\\evil.example", "https://restcoderacademy.in").href` resolves to `https://evil.example/`. Exactly as you described.

Went with the URL-parser version rather than the regex, lifted into a single `safeNextPath()` in `shared/oidc.js` that both `start.js` and `callback.js` call. Both, not one — by the time `callback.js` reads `rca_oauth_next` it is a cookie value, so it re-validates independently rather than trusting that `start.js` wrote it. The parser also covers the tab/newline/control-character variants a prefix test never sees, since it strips them before resolving.

`decodeURIComponent` in `callback.js` is guarded too. A malformed percent-escape threw there before it ever reached the filter.

**Tests:** the payload through `/start`, plus a block on `safeNextPath` itself covering backslash, double-backslash, absolute, protocol-relative, `javascript:`, control-character and non-string input. That block is the callback branch's coverage as well — it is literally the same function. 113 passing.

Took the other four as follow-ups rather than widening this PR: nonce, `MS_TENANT` issuer pinning, session `alg` check, JWKS timeout + cache. Filing them now.

**One thing worth raising before this merges:** this PR overlaps almost completely with @Pavankulli's #154 / #155 / #156. Same files — `start.js`, `callback.js`, `shared/oidc.js`, `logout.js`, `me.js` — and `schema-users.sql` against his `schema-portal.sql`. His implementation is a different design: no `next` parameter at all (role-based redirect), and it already has the nonce from your point 1. Whichever of these merges first will break the other, so it is probably worth picking a foundation before either goes in.

Abhigna1975:
Follow-ups filed: #159 (nonce) · #160 (MS_TENANT issuer pinning) · #161 (session `alg` check) · #162 (JWKS timeout + cache).

Left unassigned deliberately — #159 is already implemented in @Pavankulli's #155, so who picks these up depends on which OIDC foundation lands.

Abhigna1975:
Reworked this to remove the collision rather than wait on a decision — pushed as `5ff2e2d`.

The two sets of PRs are not alternatives. @Pavankulli's #154/#155/#156 touch **zero** React files, and #113/#114 are his tickets. This branch was carrying a backend it did not own, on top of the login screen and student home that #110/#111 actually ask for — and nothing in his PRs covers those. So the only real overlap was the backend half of this one.

**Removed from here:** `functions/auth/[provider]/start.js`, `callback.js`, `logout.js`, `me.js`, `shared/auth.js`, `shared/oidc.js`, `schema-users.sql`, `tests/portal.auth.test.js`.

His is also the better half to keep. It has the nonce from your review point 1, verifies ES256 as well as RS256, holds sessions in D1 so they can be revoked server-side rather than only expiring, reads role live from the DB instead of baking it into a JWT, and folds in the parent role from #146. The backslash open-redirect I fixed here does not exist there at all — his flow has no `next` parameter.

**What is left** is the UI: login screen, route guard, student home, and the reading of `GET /auth/me` against his response shape. That shape is the seam between two independently reviewed halves, so it is a pure function in `session-contract.js` with 10 tests pinning it — if the backend's response drifts, a test fails here instead of a student getting a blank portal.

Four states, not two: signed in, signed out, **unconfigured**, offline. Unconfigured is #42's inert-until-configured case — his backend already answers 503 when it has no secrets or no DB, which is exactly the signal the "coming soon" screen needed, so no new endpoint is required. Verified all four in a browser against his literal response shapes.

**Also fixed a layout bug the portal exposed:** `body` is a flex container and `*` sets `min-width: 0`, so `#root` sized to its content. Marketing pages are full-bleed so they always measured 100%, but the portal is a 420px card and `#root` was collapsing to it, leaving a dead band down the side of the phone. Marketing pages are pixel-identical before and after at 375 / 768 / 1440.

**Added `docs/OAUTH-SETUP.md`** — the provisioning runbook for the Google and Entra clients and the Pages secrets. That is the long-lead item for Phase 1 and it does not depend on which auth PR lands, so it can be started now.

85 tests passing, build clean. This should now merge alongside #154/#155/#156 rather than against them — though it is only useful once those land, since `/auth/me` does not exist without them.

Happy to revert if you would rather keep this branch's backend instead; it is one commit.

nikshepkulli (approved):
Re-review after the CHANGES_REQUESTED. Blocker fixed cleanly, backend duplication resolved, defence-in-depth items correctly triaged — approving.

## What's fixed since last review

- **Backslash open-redirect (the blocker) — closed in `54358a4a`.** Went with the URL-parser approach in a single `safeNextPath()` helper in `shared/oidc.js`, called from both `start.js` and `callback.js`. Callback re-validates independently rather than trusting the cookie value written by start — right call, since cookies are attacker-modifiable in transit. Tests cover backslash, double-backslash, absolute, protocol-relative, `javascript:`, control-character and non-string input. `decodeURIComponent` guarded against malformed percent-escapes. 113 tests passing.
- **Backend duplication resolved in `5ff2e2d7`.** `functions/auth/*`, `shared/auth.js`, `schema-users.sql` all dropped — clean handoff to @Pavankulli's #154/#155/#156 stack. Diff now purely frontend + docs + portal helpers, no dangling backend imports.
- **The 4 defence-in-depth items filed as follow-ups** (#159 nonce, #160 MS_TENANT issuer pinning, #161 session `alg` check, #162 JWKS timeout+cache) rather than widened into this PR. Correct call — #159 is already implemented in @Pavankulli's #155, so who owns each depends on which OIDC foundation lands.

## What's good

`PortalRoute` handles the three real states (loading / offline / anonymous) as distinct branches — "offline is not signed-out" is the single distinction the whole rural-mobile portal turns on and it's preserved. `PortalSkeleton` over spinner. Route state carries the `from` location so post-login lands the student back where they were headed. Portal chrome hidden on `/portal/*` so navbar + floating buttons don't sit on top of a student's timetable.

## Residual concerns (both small, non-blocking)

**1. PR is CONFLICTING** — rebase on main required. #138 merged today so the course files (`functions/api/portal/courses*`, `schema-courses.sql`, `shared/courses.js`, `tests/portal.courses.test.js`) that appear in your diff will drop out on rebase.

**2. Parent-role frontend route stub not added.** My heads-up comment after filing #146 asked for a `/portal/parent` placeholder route in the post-OIDC routing. Not urgent — the parent role won't be a real login target until #146's schema lands and #147's dashboard ships — but a two-line route registration in `App.jsx` + a "coming soon" component would close the routing story cleanly. Fine to file as a follow-up if you'd rather not widen the rebase.

Approving. Merge after the rebase.

Abhigna1975:
Merged `main` in and resolved the conflicts. The shape of this PR has changed a lot, so it needs a fresh look rather than a re-read of the old approval.

**What happened:** #167 landed the entire student portal on `main` while this branch sat — the auth backend (`functions/auth/*`, `shared/auth.js`, `shared/oidc.js`), the login and home screens, the route guard and the session hook. Everything this branch was proposing for #110/#111 is now a subset of what already shipped. So every one of the eight conflicted files takes `main`'s version verbatim; this branch no longer re-proposes any UI.

**Two things I dropped rather than merged:**

- **`session-contract.js` and its test are deleted.** They decode a flat `/auth/me` body — `data.id`, `data.avatar_url`, a client-side filtered `providers`. The endpoint that actually landed answers `{authenticated, providers, user: {..., picture}}`. A contract module that contradicts the endpoint is worse than not having one, because its tests would go green against a shape nothing returns. `usePortalSession` on `main` already reads the real body correctly.
- **The course APIs from this branch's tip are dropped as duplicates.** `main`'s `functions/api/portal/courses.js`, `shared/courses.js`, `schema-courses.sql` and `tests/portal.courses.test.js` are byte-identical to the ones that were here.

**What is actually left:** one file, `docs/OAUTH-SETUP.md` (+150). It is the owner-facing console walkthrough sitting behind the README's secrets table — Google consent screen and scope choice, the Entra registration, the Pages secrets, and the two things that bite (the Microsoft secret expiring, and why a native Android OAuth client is the wrong type).

I also corrected one stale claim in it while verifying it against the landed backend. It said `curl /auth/me` returns **503** when the secrets are missing. It does not — `me.js` answers **401** whether or not anything is configured, and the 503 is on `/auth/:provider/start`. The reliable signal is the `providers` array in the 401 body, and since `isConfigured()` also requires `SESSION_SECRET`, that array stays empty when only the session secret is missing — which is the failure someone setting this up is most likely to hit and least likely to diagnose.

Verified locally: 177/177 tests pass, `npm run build` clean, and the merged tree differs from `main` by exactly that one file.

This unblocks the owner-provisioned half of #42 — the code side of that issue is all on `main` now, including the Capacitor shell; what remains is someone with the Google and Microsoft accounts working through this doc.

PR #135, opened 2026-09-04
[P1] Hero — bring proof above the fold, replace mission paragraph (closes #9)

Pavankulli:
## What

Removes the 70-word generic mission paragraph. Replaces it with specific, checkable proof and a layout that works above the fold on a 375px phone.

## Changes

**`BannerContent.jsx`**
- Removed: 70-word generic mission paragraph ("we are dedicated to shaping the future of coding…")
- Added: specific subhead — placement count, named companies, city, format
- Added: compact logo strip (Wipro, Infosys, Accenture, HCL, Capgemini) — uses existing `src/assets/clients/` images, grayscale at 24px
- Added: next batch date inline with "Register Now" — driven by `useBatches()`, auto-updates when `batches.js` changes
- H1 now explicitly uses `var(--rca-fs-h1)` = 48px desktop / 32px mobile

**`Banner.css`**
- `.banner-subhead` — small semibold muted text
- `.banner-logos` — flex row, wraps on narrow viewports
- `.banner-logo` — 24px tall, grayscale, 60% opacity
- `.banner-content-btns` — flex row so button + batch date sit side by side
- `.banner-next-batch` — small semibold label

## ⚠️ Copy needs Uday's sign-off before merge

The subhead reads:
> 200+ students placed at Wipro, Infosys, Accenture, HCL & Capgemini · 4 months · offline in Bengaluru

This is **placeholder copy**, marked `TODO(copy)` in the source. Uday must confirm the number and wording. Everything else (layout, logos, batch date) is ready to ship independently.

## Test plan

- [x] Above the fold on 375px phone: headline + subhead + logos + button all visible without scrolling
- [x] Logo strip shows 5 company logos, grayscale
- [x] "Next batch · Wednesday 16 Sep" appears next to Register Now
- [x] Clicking Register Now opens enquiry modal
- [x] No old mission paragraph visible
- [x] Desktop: H1 at 48px, mobile: 32px

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #137, opened 2026-09-04
[Portal] Phase 2 — Course home + lesson list in the portal (data-light)


Abhigna1975:
Part of #43 (epic #41). Depends on the APIs in the schema ticket and the portal shell in #127.

The student-facing half of Phase 2, built for a rural India connection.

- `/portal` student home lists the courses the student is enrolled in, replacing the Phase 1 placeholder.
- `/portal/courses/:slug` shows the course with its lessons in order: title, notes, duration, and a per-lesson state (not started / in progress / done).
- **Data-light by default:** skeletons rather than spinners, no web font for the portal, covers lazy-loaded with width and height set so nothing reflows, and the lesson list renders from one request.
- Offline-friendly: the last successful course list renders from cache when the network is unavailable, with a clear "showing saved copy" line rather than an empty page.
- Empty state for a student with no enrolments that points at the marketing site, not a dead end.
- Reuses the existing portal atoms and `Portal.css` tokens from #127. No new design language.

**Out of scope, needs its own ticket:** the video player itself (low-res streaming + offline download) and instructor lesson upload. Both hinge on a storage decision that is still open, see the note on #43.

**Done when:** a signed-in student lands on /portal, sees their courses, opens one, and reads a lesson's notes on a slow connection without a blank screen.

Issue #43, opened 2026-09-01
[Portal] Phase 2 — Courses + lessons (stream low-res / offline download)


nikshepkulli:
Part of #41. Course home per enrolment; lessons as recorded video + notes, streamable low-res or **downloadable for offline**; data-light. Instructor uploads lessons (**R2** — align with epic #41; earlier Supabase reference was wrong, we're all-Cloudflare).

nikshepkulli:
Storage decision confirmed: **R2**, not Supabase. Reasons: (1) the stack is already all-Cloudflare (Pages + D1 + Workers) per epic #41; (2) R2 has zero egress, which matters most for video where bandwidth is the whole cost; (3) Supabase Storage's edge is Postgres-RLS-tied-to-Supabase-auth, and RCA doesn't use Supabase auth, so that edge is imaginary here. Body updated. Same on #44.

Abhigna1975:
Phase 2 is half done: #136 (course schema + APIs) and #137 (course home + lessons) are closed. What's left is video. Broken into build-sized tickets:

**App / frontend: @Abhigna1975**
- #188 Lesson video player on the course page: tap to load (never autoplay on a metered connection), shows the size first, resumes where the student left off
- #189 Offline download: save a lesson's video and notes to the phone, play them with no signal, show storage used, delete
- #190 Instructor upload page: pick course and lesson, upload with progress, publish. Instructors and admins only

**Backend: @Pavankulli**
- #184 R2 bucket + binding in `wrangler.toml`
- #186 Instructor-only upload API: resumable upload to R2, sets `lessons.video_url` and duration, create/edit/publish lessons
- #187 Video API: `GET /api/portal/lessons/:id/video`, session + enrolment check, range requests so seeking and resuming work
- #185 Role guard for `instructor`/`admin` (the column already exists in `users`)

Order: #184 → #186 + #185 → #187, in parallel with #188 → #189 → #190.

**Open question for @nikshepkulli:** R2 stores files but doesn't make a low-res copy. Either instructors upload a 360p export (simplest, free), or we add a transcoding step. I'd start with the 360p upload.

PR #138, opened 2026-09-04
feat(portal): course + lesson schema and session-guarded course APIs (#136)

### LINKED ISSUE #136: [Portal] Phase 2 schema + course APIs — courses, lessons, enrolment link
Part of #43 (epic #41). Depends on the session middleware in #127.

The data + API foundation for Phase 2. All-Cloudflare **D1**, matching Phase 1.

- New `schema-courses.sql`:
  - `courses`: id, slug (unique), title, summary, cover_url, published (0/1), created_at.
  - `lessons`: id, course_id, position, title, notes (markdown), video_url, duration_seconds, published, created_at. Index on `(course_id, position)`.
  - `enrolments_users`: links a portal `users.id` to a `courses.id`. The existing `enrollments` table keys off email + a course **string**, so it cannot be joined directly; this ticket adds the resolved link and a documented backfill by email.
- `GET /api/portal/courses` -> the courses the signed-in student is enrolled in. 401 without a session.
- `GET /api/portal/courses/:slug` -> one course with its published lessons in order. 404 if the student is not enrolled in it, never a 403 that confirms the course exists.
- Both go through the Phase 1 session guard. No course data is readable without a valid cookie.
- Responses are data-light: no video blobs, `cache-control: no-store`, lesson notes only on the detail call.

**Done when:** a signed-in student's enrolments resolve to real course rows, a student who is not enrolled gets a 404, and an anonymous caller gets a 401.

Abhigna1975:
Part of #43. Closes #136.

Phase 2's data foundation, stacked on #127 (the session guard it depends on).
**Base is `feat/portal-login-home`, so review #127 first.**

### Schema (`schema-courses.sql`)

- `courses` (slug unique, `published` gate) and `lessons` (ordered by `position`, indexed on `(course_id, position)`).
- `enrolments_users`, the resolved link between a portal account and a course.

The older `enrollments` table cannot serve that role. It identifies a person by
the email typed into the enquiry form and a course by a free-text string, and a
row is written before the student has ever signed in. Joining portal reads to it
would mean matching on email on every request and trusting a string to name a
course. So enrolment is resolved once into its own table; the backfill query is
documented at the foot of the schema file.

### Endpoints

| Route | Behaviour |
|---|---|
| `GET /api/portal/courses` | the caller's enrolled courses, scoped to their own user id, or 401 |
| `GET /api/portal/courses/:slug` | one enrolled course with its published lessons, or 404 |

Two decisions worth a look in review:

- **404, not 403, for a course the student is not enrolled in.** A 403 would
  confirm the course exists, so slugs could be probed. An unenrolled student and
  a nonexistent slug get a byte-identical response; there is a test asserting that.
- **503, not `[]`, when D1 is missing or the query throws.** Returning an empty
  list would tell a student who has enrolments that they have none.

Unpublished courses stay hidden even from an enrolled student, and the lesson
count only counts published lessons.

### Tests

16 new tests, 123 passing overall. Covers the 401 paths (no cookie, tampered
cookie), user-id scoping, the identical-404 property, `no-store` on per-student
responses, both 503 paths, and that a 404 reads no lesson rows.

### Not in this PR

The student-facing UI is #137. The video player and instructor upload are
unticketed, both waiting on the storage decision, see the note below.

> **Open question for @nikshepkulli:** #43 says lesson video goes to Supabase
> storage, but Phase 1 and this PR are all-Cloudflare D1 with no Supabase. R2
> would keep it on one stack. Worth settling before the upload ticket is written.

https://claude.ai/code/session_01UC2usWsvEYnGXQkZGeaA6b

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,3 +259,19 @@ npx wrangler d1 execute restcoder-enquiries --file=./schema-users.sql
 | `GET /auth/:provider/callback` | exchange the code, **verify the ID token against the provider's JWKS**, upsert into `users`, issue the session cookie |
 | `GET /auth/me` | the current user, or 401. Also reports which providers are configured. |
 | `POST /auth/logout` | clear the session cookie |
+| `GET /api/portal/courses` | the signed-in student's enrolled courses, or 401 |
+| `GET /api/portal/courses/:slug` | one enrolled course with its published lessons. **404 when the student is not enrolled**, so slugs cannot be probed. |
+
+### Course content (Phase 2)
+
+Courses and lessons live in the same D1 database as `users`:
+
+```
+npx wrangler d1 execute restcoder-enquiries --file=./schema-courses.sql
+```
+
+`enrolments_users` is the resolved link between a portal account and a course.
+The older `enrollments` table cannot serve that role: it identifies a person by
+the email typed into the enquiry form and a course by free text, and its rows
+predate any sign-in. The backfill that maps one to the other is at the foot of
+`schema-courses.sql`.
--- a/functions/api/portal/courses.js
+++ b/functions/api/portal/courses.js
@@ -1,0 +1,29 @@
+// GET /api/portal/courses — the signed-in student's enrolled courses (#136).
+// Session-guarded: no cookie, no course data. Never cached, because the
+// response is specific to one student.
+import { getSession } from "../../../shared/auth.js";
+import { listEnrolledCourses } from "../../../shared/courses.js";
+
+const json = (body, status = 200) =>
+  new Response(JSON.stringify(body), {
+    status,
+    headers: { "content-type": "application/json", "cache-control": "no-store" },
+  });
+
+export async function onRequestGet(context) {
+  const { request, env } = context;
+
+  const session = await getSession(request, env);
+  if (!session) return json({ error: "unauthenticated" }, 401);
+
+  // No D1 binding is a server fault, not an empty course list. Returning []
+  // here would tell a student who has enrolments that they have none.
+  if (!env.DB) return json({ error: "unavailable" }, 503);
+
+  try {
+    const courses = await listEnrolledCourses(env.DB, session.uid);
+    return json({ courses });
+  } catch {
+    return json({ error: "unavailable" }, 503);
+  }
+}
--- a/functions/api/portal/courses/[slug].js
+++ b/functions/api/portal/courses/[slug].js
@@ -1,0 +1,32 @@
+// GET /api/portal/courses/:slug — one enrolled course with its lessons (#136).
+//
+// A student who is not enrolled gets the same 404 as a slug that does not
+// exist. A 403 would confirm the course is real, which is not something an
+// outsider should be able to learn by guessing slugs.
+import { getSession } from "../../../../shared/auth.js";
+import { getEnrolledCourse } from "../../../../shared/courses.js";
+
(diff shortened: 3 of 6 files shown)


nikshepkulli (approved):
**Approve.** Small, focused, well-designed. The decisions in the PR body are all correct:

- **404-not-403 for unenrolled** is right, and the test asserts byte-identical responses ✅
- **503-not-`[]` when D1 is down** avoids the "you have no enrolments" lie ✅
- **`enrolments_users` as a resolved link** rather than joining on `enrollments.email` is the right call — matching a portal identity to an enquiry-form string on every request would be both slow and wrong
- All SQL uses `?1` / `?2` parameter binding (no injection surface) ✅
- Cascade behaviour on `user_id` and `course_id` is correct ✅
- Cache-control `no-store` on per-student responses ✅

Test coverage is genuinely good: 401 (no cookie / tampered cookie), user-id scoping (asserted via query arg log), 503 on missing DB and on thrown query, 404 for unenrolled, 404 that reads no lesson rows, 404 byte-identical to nonexistent-slug.

### 🟡 Two test gaps worth closing (nice-to-have, not blocking)

The PR body promises two behaviours that only exist in the SQL WHERE clauses, and the current tests assert them by string-matching the query text — a refactor to a different SQL shape would let the guarantee lapse silently.

**1. Behavioural test: unpublished course withheld from an enrolled student.**

\`\`\`js
it("404s when the enrolled course is unpublished", async () => {
  const db = fakeDB([["FROM enrolments_users", []]]);  // no row matches published=1
  const res = await detailGet({
    request: req("https://x/a", await session()),
    env: { SESSION_SECRET: SECRET, DB: db },
    params: { slug: "draft-course" },
  });
  expect(res.status).toBe(404);
});
\`\`\`

**2. Behavioural test: unpublished lessons filtered from course detail.**

Right now the assertion is \`expect(log[1].sql).toContain("ORDER BY position ASC")\`. Add:

\`\`\`js
it("only returns published lessons on a course detail", async () => {
  const log = [];
  await getEnrolledCourse(
    fakeDB([["FROM enrolments_users", [{ id: 5, slug: "fde" }]], ["FROM lessons", []]], log),
    1, "fde",
  );
  expect(log[1].sql).toContain("published = 1");
});
\`\`\`

Both are one-line additions; both catch a real regression class (someone refactors the query and drops the filter).

### ⚪ Schema follow-ups (not this PR)

- **\`enrolments_users\` naming** — British "enrolments" vs the older American "enrollments" table. Comment justifies it (deliberately separate table), but the two spellings living side-by-side will trip someone up. Worth a ticket to rename one when the old \`enrollments\` table can be retired.
- **No \`status\` / \`revoked_at\` on \`enrolments_users\`** — revocation today means DELETE, which loses history. Add when admin controls arrive; YAGNI for Phase 2.

### Answer to the open question in the PR body

> "#43 says lesson video goes to Supabase storage… R2 would keep it on one stack. Worth settling before the upload ticket is written."

**R2.** You're already on Cloudflare (Pages Functions + D1), R2 has zero egress fees (video hurts otherwise), Workers can mint signed URLs per session for auth-gated access, and pulling in Supabase for one table's worth of blobs makes future ops harder without adding capability the portal needs. Supabase makes sense if you specifically want its Postgres + Row-Level-Security + realtime story — the portal doesn't. Update #43 to say R2, close #135 as the storage decision.

PR #139, opened 2026-09-04
feat(portal): course home and lesson list for the student portal (#137)

### LINKED ISSUE #137: [Portal] Phase 2 — Course home + lesson list in the portal (data-light)

Abhigna1975:
Part of #43. Closes #137.

The student-facing half of Phase 2, stacked on #138 (the APIs it calls).
**Base is `feat/portal-courses-api-136`, so review #127 then #138 first.**

### What a student sees

- `/portal` "My courses" is now their real enrolments, replacing the Phase 1 placeholder.
- `/portal/courses/:slug` is one course with its lessons in order; each lesson expands in place to show its notes.

### Built for the connection these students have

- **The saved copy paints first.** The course list renders from `localStorage`
  before the request goes out, so a slow network costs a stale label ("Showing
  your saved copy") rather than a blank screen. The cache is keyed per user id,
  because a shared phone would otherwise show one sibling the other's courses,
  and it ages out after 7 days so it is never mistaken for the truth.
- **A failed request is "you are offline", never "you have no courses."**
  Telling an enrolled student they have nothing is the one answer that is worse
  than an error, so the empty state and the network state are kept distinct.
- Skeletons rather than spinners, no photographs, no extra web font, and the
  list renders from a single request.
- Every tappable row is at least 48px; lesson numbers and durations use tabular
  figures so lesson 9 to 10 does not shift the title.

### Two deliberate omissions

- **No video player.** #43 has not settled where lesson video lives, so each
  lesson says so in a line of text rather than offering a play button that does
  nothing. Needs its own ticket once the storage call is made.
- **Notes render as text, not parsed markdown.** No parser shipped to the
  client and no untrusted HTML, which seemed the right default until someone
  actually needs formatting.

### Tests

24 new tests, 142 passing overall. This repo has no jsdom or testing-library
and does not test components, so the cache rules and duration formatting live
in `coursesCache.js` and are tested in node the way `batchDateUtils` is: per
user keying, expiry including the exact boundary, corrupt and wrong-shape
entries, storage that throws, and the formatter's edge values. The components
stay thin enough to read.

`npm run build` passes.

> **Still open for @nikshepkulli:** lesson video storage. #43 says Supabase,
> but Phase 1 and #138 are all-Cloudflare D1. R2 would keep it on one stack.

https://claude.ai/code/session_01UC2usWsvEYnGXQkZGeaA6b

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,7 +212,8 @@ app ships to students on rural connections.
 
 ## Student portal (Phase 1)
 
-`/portal/login` and `/portal` are the student portal (#110, #111). They are
+`/portal/login`, `/portal` and `/portal/courses/:slug` are the student portal
+(#110, #111, #137). They are
 **inert until configured**: with no OAuth secrets set, `/auth/me` reports no
 providers and the login screen says "coming soon" rather than rendering buttons
 that lead to a provider error page. The marketing site is unaffected either way.
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -16,6 +16,7 @@ import ScrollToTop from "./components/ScrollToTop";
 import PortalRoute from "./components/portal/PortalRoute";
 import PortalLogin from "./components/portal/PortalLogin";
 import PortalHome from "./components/portal/PortalHome";
+import PortalCourse from "./components/portal/PortalCourse";
 import Modal from 'react-modal';
 import EnquiryForm from './components/forms/Enquiry Form/EnquiryForm';
 import EnrollForm from './components/forms/EnrollForm/EnrollForm';
@@ -119,6 +120,16 @@ function App() {
           element={
             <PortalRoute>
               {({ user, logout }) => <PortalHome user={user} logout={logout} />}
+            </PortalRoute>
+          }
+        />
+        {/* One course and its lessons (#137). Behind the same guard, so a
+            direct link from outside lands on login and returns here after. */}
+        <Route
+          path="/portal/courses/:slug"
+          element={
+            <PortalRoute>
+              <PortalCourse />
             </PortalRoute>
           }
         />
--- a/src/components/portal/CourseList.jsx
+++ b/src/components/portal/CourseList.jsx
@@ -1,0 +1,84 @@
+import PropTypes from "prop-types";
+import { Link } from "react-router-dom";
+import "./Portal.css";
+
+/**
+ * The "My courses" section on /portal (#137).
+ *
+ * Text only. A course cover is a photograph on a metered connection, so the
+ * list renders from the one request it already made and never blocks on an
+ * image; the tile carries the title, a one-line summary and a lesson count.
+ */
+function CourseList({ status, courses, cached, reload }) {
+  if (status === "loading") {
+    return (
+      <div className="portal-skeleton" aria-busy="true">
+        <div className="sk sk-card" />
+        <div className="sk sk-card" />
+      </div>
+    );
+  }
+
+  if (status === "offline" || status === "error") {
+    return (
+      <div className="portal-card portal-card--quiet" role="status">
+        <p className="portal-card-title">
+          {status === "offline" ? "You are offline" : "We could not load your courses"}
+        </p>
+        <p className="portal-card-sub">
+          {status === "offline"
+            ? "Your courses will appear here once you are back on a connection."
+            : "This is on our side, not yours. Please try again in a moment."}
+        </p>
+        <button type="button" className="portal-btn portal-btn--ghost" onClick={reload}>
+          Try again
+        </button>
+      </div>
+    );
+  }
+
+  if (!courses.length) {
+    // Not a dead end: a student with no enrolment is told where to go next.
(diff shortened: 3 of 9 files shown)


nikshepkulli (approved):
**Approve.** Nice UI work — the data-light decisions in the PR body are all defensible, the cache is unit-testable rather than tested through a jsdom component, and the accessibility work is real (`aria-busy` on skeletons, `aria-expanded` on the lesson toggle, `aria-labelledby` on sections, `aria-current` / `aria-disabled` on the tab bar, initials-avatar with `aria-hidden`). The "cached-is-labelled" pattern is the right call — silently showing 5-day-old data is exactly how portals lose trust.

A couple of specific things I checked:

- **`formatDuration` edges** — NaN / null / undefined / negative all `→ "—"`, 30s rounds to "1m", 3599s rolls to "1h", 3600s exact is "1h". Good.
- **XSS surface** — lesson `notes` render as `{l.notes}` (React auto-escapes), `Link to={/portal/courses/${c.slug}}` treats value as a router path so a malicious slug would not become an open redirect. Fine.
- **Slug encoding** — `encodeURIComponent(slug)` before the fetch. Fine.
- **`useOnline`** — the "`navigator.onLine` is only trustworthy when it says false" comment is right and rare to see stated.
- **`PortalCourse.useEffect`** cancels correctly via the `let live = true; return () => { live = false; }` pattern.

### 🟡 Two small defence-in-depth items (nice-to-have, not blocking)

**1. `useCourses` should match `PortalCourse`'s cancel pattern.**

Right now the fetch has no cancel/staleness check, so if `userId` changes during a fetch (or the component unmounts on a navigate-away mid-load), a late resolve calls `setState` on an unmounted component / applies user A's data to user B. React 18 suppresses the warning but the underlying race is real.

\`\`\`js
const load = useCallback(async () => {
  let live = true;
  const saved = readCache(userId);
  if (saved) { setCourses(saved.courses); setCached(true); setStatus("ready"); }
  else setStatus("loading");
  try {
    const res = await fetch("/api/portal/courses", { credentials: "same-origin" });
    if (!live) return;
    // ...rest identical
  } catch {
    if (live && !saved) setStatus("offline");
  }
  return () => { live = false; };
}, [userId]);
\`\`\`

(Or wrap it in an AbortController the effect cleanup aborts — same effect, marginally cleaner.)

**2. `logout` should clear this user's cache.**

`usePortalSession.logout` (from #127) only clears the session cookie. The course-list localStorage stays keyed to that `uid`. Not a security issue — the cached data is course titles + summaries + lesson counts, not PII or payment — but on a shared phone, if the same student signs back in they see pre-logout data before the fresh fetch completes. Two lines:

\`\`\`js
import { clearCache } from "./coursesCache";
// in logout, before setUser(null):
if (user?.id) clearCache(user.id);
\`\`\`

Belongs in `usePortalSession` since that's where logout lives, not this PR strictly — but flagging it here because the responsibility only exists once `coursesCache` does.

### ⚪ Out-of-scope for this PR, worth a follow-up ticket

- **`white-space: pre-wrap` on `.portal-lesson-notes`** — respects the instructor's line breaks in the D1 `notes` field without needing a markdown parser. One CSS line; delivers most of what "markdown notes" would give a student without any parsing/security surface.
- **Cache the currently-viewed course too.** The list on `/portal` survives offline, but tapping into a course and losing connection mid-lesson drops everything. Same key pattern would extend to `rca_portal_course:<uid>:<slug>`.
- **Deep-link to an open lesson.** Currently `openId` is component state; a `?lesson=<id>` query param would let a student share a URL back into where they were.

Ship it.

nikshepkulli:
Auto-closed by GitHub when I merged #138 with `--delete-branch` and the base branch (`feat/portal-courses-api-136`) went with it. Sorry @Abhigna1975 — should have caught the stacking. Recreated as a new PR against main; work is preserved.

Issue #147, opened 2026-09-06
[Portal] Placement-readiness score for parents — composite of attendance + assignments + gradebook + projects


nikshepkulli:
## Why this exists

Parents don't want to read four separate dashboards — attendance %, assignment completion %, gradebook average, project submission count. They want **one number** that answers "is my kid on track to get placed?" Placement readiness is the synthesized metric that composes those inputs into a single, honest score.

This is the actual content of the parent dashboard route (`/portal/parent`) that #146 stubs out.

### 1. The score — what goes into it

A weighted composite, computed server-side, from the four Phase 3–5 data sources:

| Input | Weight | Data source |
|---|---|---|
| **Attendance** — % of scheduled sessions attended | 25% | #46 |
| **Assignment completion** — % of assigned work submitted on time | 30% | #44 |
| **Gradebook average** — instructor-scored work, normalized to 0-100 | 30% | #45 |
| **Project submission** — count of projects submitted vs cohort median | 15% | #44 (project subtype) |

Score = weighted sum → 0–100 integer. No decimals shown to parents. No colour-coding (green/amber/red) — see "Design principles" below.

Weights aren't final — file a follow-up ticket to re-tune once we have 4-6 weeks of data. Starting numbers reflect "attendance and consistency matter, one bad exam doesn't sink you."

### 2. What the parent sees

`/portal/parent` for a logged-in parent linked to at least one student:

- Child's name + course + batch start date
- **Big number: placement readiness score (0-100)**
- Below it, in smaller type: the four inputs with their own percentages, so a parent who wants to drill down can see WHERE the score came from
- Trend arrow: score change over last 2 weeks (up / flat / down). Only shown after 2 weeks of data.
- Last updated timestamp
- No CTA, no "message the instructor" button in v1 — parents in rural India are more likely to phone Uday directly, so don't build a chat surface until we see a real need

Multi-child parents: list all linked students; parent picks one to view. Same layout per child.

### 3. What the parent does NOT see

- Individual grades or comments on specific assignments — that's the student's own portal
- Peer comparison ("your child ranks 4th of 12") — creates the wrong incentives, and rural parents may act on ranking data in ways that hurt the student
- Predictive placement probability — we're not making the claim "this score means Ashish will get placed", we're saying "here's how consistent your child has been"
- Any monetary or upsell messaging — this is a trust surface, not a sales one

### 4. Compute cadence

[...]

Issue #114, opened 2026-09-02
[Portal] Signed session cookie + /auth/me and /auth/logout


Abhigna1975:
Part of #42 (epic #41).

How the app knows who is logged in, on web and inside the Capacitor shell.

- Issue a **JWT signed with Web Crypto** (`SESSION_SECRET`) as an **HttpOnly, Secure, SameSite** cookie on successful callback; sensible expiry + a documented refresh/renewal rule.
- `GET /auth/me` → the current user (id, name, email, avatar, role) or 401.
- `POST /auth/logout` → clear the cookie (and revoke the session row if we keep one).
- Middleware that guards portal API routes and rejects a missing/expired/tampered token.
- **Native shell:** confirm the cookie survives the deep-link return in Capacitor; fall back to a documented token hand-off if the WebView will not keep it.
- Tests for: valid token, expired token, wrong signature, missing cookie.

**Done when:** `/auth/me` returns the signed-in student on web and in the Android app, and logout genuinely ends the session.

PR #154, opened 2026-09-07
feat(portal): D1 schema + upsert helper + admin UI (closes #112, closes #146)

### LINKED ISSUE #112: [Portal] D1 schema for the portal — users, roles, sessions

### LINKED ISSUE #146: [Portal] Add 'parent' role + parent↔student linking in D1 schema (BLOCKS #112)
## Why this exists

Ticket #112 as currently written assumes only three roles (student, instructor, admin) and no way for a parent to log in and see their own child's data. That contradicts the epic's "access is the product" thesis for rural India, where the parent — not the student — is typically the one making the enrolment decision AND paying, and needs visibility into whether the child is showing up and progressing.

We can't retrofit this after the schema ships without a real migration. Filing this to fold into #112 BEFORE it merges, not after.

### 1. `users.role` extended

Add `parent` to the allowed role set in `schema-portal.sql`. In the CHECK constraint (or the app-level validator, whichever pattern you go with):

- `student` (default) — existing
- `instructor` — existing
- `admin` — existing
- `parent` — **new**

### 2. New table `parent_students`

Links a parent user to one or more student users. A parent can have multiple children enrolled; a student can have zero or more parents (some students are self-enrolled adults).

- `id` (uuid, pk)
- `parent_user_id` (fk → `users.id`)
- `student_user_id` (fk → `users.id`)
- `relation` (text, e.g. 'mother', 'father', 'guardian') — optional, for display only
- `created_at`
- Unique index on `(parent_user_id, student_user_id)`

RLS or app-level: a parent can only read student data where `parent_students.parent_user_id = current_user.id`. Never expose bulk student data to a parent user.

### 3. Provisioning flow

Two ways a parent-student link gets created — pick whichever is simpler for now, defer the other:

**Option A (simpler):** admin-created link. Admin logs into `/portal/admin/users`, finds the student, adds the parent's email/name, system creates or upserts the parent user row and inserts the `parent_students` link. Parent then signs in with Google/MS SSO against that email and lands on their linked child's dashboard.

**Option B (later):** student-invited link. Student sends parent an invite email with a signed token; parent clicks, SSOs in, link auto-created.

Do A now. B is fine as a follow-up ticket after we have parents actually using it.

### 4. Session cookie already covers this

No change needed to #114 — the session cookie carries `user_id` + `role`, and downstream API handlers use the role for gating. Parent-scoped data reads happen via the role check + the `parent_students` join.

### 5. Migration seed data

Zero seed. No hardcoded parent-student links. Admin creates them as real parents sign up.

## Where the parent lands after login

After OIDC callback, the session-created flow branches on role:
- student → `/portal/home` (existing behavior in #111)
- instructor → `/portal/teach` (deferred, not scoped here)
- admin → `/portal/admin`
- parent → `/portal/parent` (new placeholder route — render "coming soon" content until #147 ships the actual dashboard)

The parent dashboard content itself is deferred to ticket **#147** (placement-readiness composite). This ticket only sets up the data model + role + route stub.

## Non-goals

- No parent-view dashboard content — that's #147
- No parent-invited-by-student flow — future ticket if useful
- No multi-parent conflict resolution (both parents editing settings, etc.) — parents only READ, don't WRITE
- No parent notifications / weekly digest emails — future ticket
- No self-enrolled parents (parent signing up without an admin linking a child) — those users see an empty state

## Done when

- [ ] `parent` role accepted by the schema constraint
- [ ] `parent_students` table exists with unique-index on (parent, student) pair
- [ ] Admin UI at `/portal/admin/users` can create parent users and link them to a student
- [ ] A parent SSO'd in with a linked student is redirected to `/portal/parent` after login
- [ ] A parent trying to fetch any student's data other than their linked child gets a 403 (RLS or app-level check)
- [ ] The migration applies cleanly on a fresh D1 AND on top of #112 in progress

## Blocks / blocked-by

- **Blocks:** #147 (placement-readiness dashboard — depends on parent role existing)
- **Blocked by:** nothing — this needs to be folded INTO #112 before it merges, not sequenced after
- **Related to:** #114 (session cookie already gates by role — no change), #111 (student home doesn't change)

## Parent epic

#41

@Pavankulli — filing this ahead of your #112 merge specifically so the parent role can land as part of the same schema PR. Adding it later means a data migration + auth rework, which is worse than a 10-line schema addition now.

Pavankulli:
## What

Lays the complete data foundation for the student portal. #146 is folded into #112 as a single migration so the parent role ships from day one and never needs a retrofitted ALTER TABLE.

## Files

| File | Purpose |
|---|---|
| `schema-portal.sql` | Three new tables: `portal_users`, `portal_sessions`, `parent_students` |
| `shared/portalDb.js` | `upsertPortalUser`, `createPortalSession`, `resolvePortalSession`, `revokePortalSession` |
| `functions/admin/portal-users.js` | Basic-auth admin UI at `/admin/portal-users` |
| `functions/portal/parent.js` | Stub `/portal/parent` — "coming soon" until #147 ships |
| `wrangler.toml` | Inline migration commands (no binding changes) |

## Schema highlights

**`portal_users`**
- Roles: `student` (default) · `instructor` · `admin` · `parent` — all four in the initial CHECK constraint, no ALTER later
- `provider` nullable → admin can pre-create a parent row with just email+name; the upsert fills in provider+subject on first SSO login
- Partial unique index on `(provider, provider_subject) WHERE provider IS NOT NULL` — enforces no duplicate SSO accounts while allowing pre-created rows

**`portal_sessions`**
- Explicit rows (not purely stateless signed cookies) so individual sessions can be revoked server-side without rotating the signing secret
- `revoked_at IS NULL` = active; set on logout (#114)

**`parent_students`**
- `(parent_user_id, student_user_id)` unique index — no duplicate links
- `ON DELETE CASCADE` on both FKs — deleting a user cleans up their links automatically

## upsert logic (`shared/portalDb.js`)

Three-case upsert called by the OIDC callback (#113):
1. Match by `(provider, provider_subject)` → update name/email/avatar + last_login_at
2. Match by email where `provider IS NULL` (admin pre-created parent) → fill provider columns, preserve role=parent
3. No match → insert new user with role=student

## Admin UI (`/admin/portal-users`)

- Lists all portal users with role, provider, last login
- Pre-create parent accounts (email + name → role=parent, matched by email at first SSO)
- Link parent → student, with role validation
- Unlink and delete-user with confirm dialogs

## Apply the migration

```bash
# Local dev
npx wrangler d1 execute restcoder-enquiries --local --file=schema-portal.sql

# Production
npx wrangler d1 execute restcoder-enquiries --remote --file=schema-portal.sql
```

## Done when (from issue)

- [x] `parent` role accepted by the schema CHECK constraint
- [x] `parent_students` table with unique index on (parent, student) pair
- [x] Admin UI can create parent users and link them to a student
- [x] Migration applies cleanly on a fresh D1 (idempotent — all `CREATE TABLE IF NOT EXISTS`)
- [ ] Parent SSO'd in → redirected to `/portal/parent` — depends on #113 (OIDC endpoints)
- [ ] Parent fetching other student's data → 403 — depends on #114 (session + auth/me)

## Unlocks

#113 (OIDC endpoints) and #114 (session cookie) can now be built on top of this schema.

## Test plan

- [ ] Apply `schema-portal.sql` locally — no errors
- [ ] Re-apply — no errors (idempotent)
- [ ] Visit `/admin/portal-users` with correct credentials → page loads, shows empty tables
- [ ] Pre-create a parent: fill email + name → row appears in user list with role=parent
- [ ] Attempt to create duplicate email → error message shown
- [ ] Visit `/portal/parent` → "coming soon" stub renders

closes #112
closes #146

### DIFF
--- a/functions/admin/portal-users.js
+++ b/functions/admin/portal-users.js
@@ -91,10 +91,15 @@ export async function onRequestPost(context) {
 
     const id = crypto.randomUUID();
     const now = new Date().toISOString();
-    await env.DB.prepare(
-      "INSERT INTO portal_users (id, provider, provider_subject, email, name, role, created_at, last_login_at) " +
-        "VALUES (?1, NULL, NULL, ?2, ?3, 'parent', ?4, ?4)"
-    ).bind(id, email, name, now).run();
+    try {
+      await env.DB.prepare(
+        "INSERT INTO portal_users (id, provider, provider_subject, email, name, role, created_at, last_login_at) " +
+          "VALUES (?1, NULL, NULL, ?2, ?3, 'parent', ?4, ?4)"
+      ).bind(id, email, name, now).run();
+    } catch {
+      // UNIQUE constraint on email — can happen on a race between two admin tabs
+      return redirect(request, "err", `Could not create account for ${email} — email may already exist.`);
+    }
     return redirect(request, "ok", `Parent account created for ${email}. They will receive role=parent on first SSO login.`);
   }
 
@@ -125,9 +130,13 @@ export async function onRequestPost(context) {
     if (dupe) return redirect(request, "err", "This parent↔student link already exists.");
 
     const id = crypto.randomUUID();
-    await env.DB.prepare(
-      "INSERT INTO parent_students (id, parent_user_id, student_user_id, relation, created_at) VALUES (?1, ?2, ?3, ?4, ?5)"
-    ).bind(id, parentId, studentId, relation || null, new Date().toISOString()).run();
+    try {
+      await env.DB.prepare(
+        "INSERT INTO parent_students (id, parent_user_id, student_user_id, relation, created_at) VALUES (?1, ?2, ?3, ?4, ?5)"
+      ).bind(id, parentId, studentId, relation || null, new Date().toISOString()).run();
+    } catch {
+      return redirect(request, "err", "Could not create link — it may already exist.");
+    }
     return redirect(request, "ok", "Parent↔student link created.");
   }
 
@@ -135,7 +144,8 @@ export async function onRequestPost(context) {
   if (action === "unlink") {
     const id = String(form.get("id") || "").trim();
     if (!id) return redirect(request, "err", "Missing link id.");
-    await env.DB.prepare("DELETE FROM parent_students WHERE id = ?1").bind(id).run();
+    const res = await env.DB.prepare("DELETE FROM parent_students WHERE id = ?1").bind(id).run();
+    if (!res.meta?.changes) return redirect(request, "err", "Link not found.");
     return redirect(request, "ok", "Link removed.");
   }
 
@@ -143,7 +153,8 @@ export async function onRequestPost(context) {
   if (action === "delete-user") {
     const id = String(form.get("id") || "").trim();
     if (!id) return redirect(request, "err", "Missing user id.");
-    await env.DB.prepare("DELETE FROM portal_users WHERE id = ?1").bind(id).run();
+    const res = await env.DB.prepare("DELETE FROM portal_users WHERE id = ?1").bind(id).run();
+    if (!res.meta?.changes) return redirect(request, "err", "User not found.");
     return redirect(request, "ok", "User removed.");
   }
 
--- a/functions/admin/portal-users.js
+++ b/functions/admin/portal-users.js
@@ -1,0 +1,296 @@
+// /admin/portal-users — manage portal users and parent↔student links.
+// Same Basic Auth as the rest of the admin panel (ADMIN_PASSWORD secret).
+//
+// Actions (POST form):
+//   add-parent  — pre-create a parent user row so the parent gets role=parent
+//                 when they SSO in for the first time (matched by email).
+//   link        — create a parent_students row linking two existing users.
+//   unlink      — remove a parent_students link (id = parent_students.id).
+//   delete-user — remove a portal_users row (cascades to sessions + links).
+//
+// Part of #112 + #146.
+import { escapeHtml, requireAdminAuth } from "../../shared/serverUtil.js";
+
+function html(body, status = 200) {
+  return new Response(body, {
+    status,
+    headers: { "Content-Type": "text/html; charset=utf-8" },
+  });
+}
+
+function redirect(request, param, msg) {
+  const url = new URL(request.url);
+  url.search = "";
+  url.searchParams.set(param, msg);
+  return Response.redirect(url.toString(), 303);
+}
+
+// ---------------------------------------------------------------------------
+// GET — render the admin page
+// ---------------------------------------------------------------------------
+export async function onRequestGet(context) {
+  const { request, env } = context;
+  const auth = requireAdminAuth(request, env);
+  if (auth) return auth;
+  if (!env.DB) return html(page([], [], "Storage not configured.", null), 500);
+
+  let users = [], links = [];
+  try {
+    const usersRes = await env.DB.prepare(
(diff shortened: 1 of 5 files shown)


Pavankulli:
## Self-review (Pavankulli)

Reviewed `schema-portal.sql`, `shared/portalDb.js`, `functions/admin/portal-users.js`, `functions/portal/parent.js`.

### Issues found and fixed (commit 68fa96b)

| Location | Finding | Fix |
|---|---|---|
| `portal-users.js` add-parent | INSERT had no error handling — concurrent admin tabs could both pass the email uniqueness SELECT and then one INSERT would throw uncaught | Wrapped in try/catch; UNIQUE index on email is last line of defence |
| `portal-users.js` link | `parent_students` INSERT had no error handling | Same try/catch — catches race-condition constraint violation |
| `portal-users.js` unlink | DELETE returned "Link removed" even if `id` didn't exist | Check `res.meta.changes`; return error if 0 rows affected |
| `portal-users.js` delete-user | Same silent success on missing id | Same fix |

### False positive from reviewer (not changed)

`resolvePortalSession` expiry check: `expires_at > ?2` is correct — if `expires_at` equals `now` exactly, the session IS expired and the condition is FALSE → rejected. Changing to `>=` would allow a session to pass at the instant of expiry.

### No remaining issues

Schema constraints, indexes, cascade deletes, XSS escaping, auth gating all correct.

nikshepkulli (approved):
Approving. Foundation lands correctly and #146 is genuinely folded into #112 (parent role in the initial CHECK constraint, `parent_students` table with the unique (parent, student) pair index, ON DELETE CASCADE from `portal_users` → sessions + links). Clean.

**Notable design choices that land right:**

- **Partial unique index on `(provider, provider_subject) WHERE provider IS NOT NULL`** — clever. Lets admin pre-create parent rows without a provider so their first SSO fills it in (`upsertPortalUser` Case 2), without needing a nullable-column-in-unique-index workaround.
- **Email uniqueness enforced.** Prevents two Google/MS accounts sharing an email from creating duplicate portal users.
- **`portal_sessions` explicit rows, not stateless signed cookies.** Server-side revocation without rotating the signing secret — matches your comment. Right call for a portal where a compromised device needs to be logged out immediately.
- **Admin UI: no privilege escalation vector.** `add-parent` hardcodes `role='parent'`; there's no form field or action for granting `admin`. Delete-user cascades correctly. All action handlers re-check role (parent must have role=parent before being linkable). SQL is fully parameterised; no injection surface.

**Two minor observations (non-blocking):**

- Email stored plaintext. Fine for a school; would need encryption at rest if this data set expanded to financial/medical. Note for the future.
- `parent_students.relation` is TEXT with no CHECK constraint — an admin could type anything. Not a security issue because parents only READ, and it's display-only per your comment. Acceptable.

**Merge order:** land #154 first, then rebase #155 on it, then #156 on that. Otherwise the trio ships three copies of `shared/portalDb.js` and `shared/oidc.js` and future edits diverge silently.

PR #155, opened 2026-09-07
feat(auth): Google + Microsoft OIDC — Authorization Code + PKCE (closes #113)

### LINKED ISSUE #113: [Portal] Google + Microsoft OIDC endpoints (Authorization Code + PKCE)

Pavankulli:
## What

Hand-rolled OIDC in Cloudflare Pages Functions. No auth SaaS, no Supabase.

Depends on #154 (portal D1 schema) being applied first — the `portal_users` and `portal_sessions` tables must exist before the callback can write to them.

### `GET /auth/:provider/start`

Initiates the login flow for `google` or `microsoft`.

1. Validates provider + secrets configured (inert-until-configured — 503 if any secret absent)
2. Generates `code_verifier` (32 random bytes → base64url), `code_challenge` (base64url(SHA-256(verifier))), random `state`, random `nonce`
3. Signs `{state, codeVerifier, nonce, provider, redirectUri, exp}` into a 10-minute `HttpOnly; Secure; SameSite=Lax` cookie using HMAC-SHA256 (`COOKIE_SECRET` env var)
4. Redirects browser to provider's authorization endpoint with PKCE params + `prompt=select_account`

### `GET /auth/:provider/callback`

Handles the provider redirect back after user consent.

1. Validates signed state cookie (HMAC, expiry, provider field)
2. Validates `state` query param matches cookie
3. **Token exchange server-side** — `POST` to provider's token endpoint with `code_verifier` (PKCE proof), `client_secret` — code never touches the browser
4. **JWKS verification** — fetches provider JWKS (CF-edge cached), finds key by `kid`, imports via `crypto.subtle`, verifies RS256/ES256 signature
5. Validates claims: `exp`, `iat`, `aud` (exact), `iss` (exact for Google, prefix for Microsoft), `nonce` (replay protection)
6. Upserts user in D1 via `upsertPortalUser` (three-case logic from #154)
7. Creates 7-day session via `createPortalSession`
8. Clears OIDC state cookie, sets `__session` session cookie
9. Redirects to role-appropriate portal page

## Files

| File | Purpose |
|---|---|
| `shared/oidc.js` | Pure helpers: PKCE, state cookie sign/verify, JWKS + ID token verification |
| `shared/oidc.test.js` | 14 tests incl. RFC 7636 Appendix B PKCE test vector |
| `functions/auth/[provider]/start.js` | Start endpoint |
| `functions/auth/[provider]/callback.js` | Callback endpoint |

## Security properties

- PKCE (S256) prevents authorization code interception attacks
- State parameter prevents CSRF
- Nonce prevents ID token replay attacks
- JWKS signature verification — never trusts an unsigned/self-signed token
- `code_verifier` and `client_secret` never leave the server
- State cookie HttpOnly + SameSite=Lax — not readable by JS, CSRF-resistant
- Session cookie HttpOnly + Secure + SameSite=Lax
- Inert-until-configured — missing secrets → 503, never 500

## Env vars required

| Var | What |
|---|---|
| `GOOGLE_CLIENT_ID` | Google OAuth 2.0 Client ID |
| `GOOGLE_CLIENT_SECRET` | Google OAuth 2.0 Client Secret |
| `MS_CLIENT_ID` | Microsoft Entra App Registration Client ID |
| `MS_CLIENT_SECRET` | Microsoft Entra Client Secret |
| `COOKIE_SECRET` | Random 32+ char secret for signing state cookies |

Owner-provisioned (not in this PR): Google Cloud OAuth client + Microsoft Entra app registration. Both must register the redirect URI `https://restcoderacademy.in/auth/{provider}/callback` (and `http://localhost:8788/auth/{provider}/callback` for local dev).

## Not in this PR

- `/auth/me` and `/auth/logout` → #114
- Capacitor deep-link redirect URI → when mobile app ships

## Tests

89/89 passing (14 new OIDC tests).

## Test plan (requires real OAuth credentials)

- [ ] `GET /auth/google/start` → redirects to Google consent page
- [ ] Complete Google login → redirects to `/portal/home` (student role)
- [ ] `GET /auth/microsoft/start` → redirects to Microsoft consent page
- [ ] Complete Microsoft login → redirects to `/portal/home`
- [ ] After login, `portal_users` row exists in D1 with correct email/name/role
- [ ] Second login with same account → upserts, does not create duplicate row
- [ ] Admin pre-creates parent in `/admin/portal-users` → parent SSOs in → role=parent, redirects to `/portal/parent`
- [ ] Missing `GOOGLE_CLIENT_ID` → `GET /auth/google/start` returns 503 JSON

closes #113


Pavankulli:
## Self-review (Pavankulli)

Security-focused review of `shared/oidc.js`, `start.js`, `callback.js`.

### Issues found and fixed (commit 261f518)

| Location | Finding | Fix |
|---|---|---|
| `oidc.js` JWKS key selection | No `kid` in token header → picked first `use=sig` key. If JWKS had multiple signing keys without `kid`, wrong key could be chosen → verification could fail or accept wrong token | Now requires exactly one signing key when token has no `kid`; throws `"cannot select unambiguously"` otherwise |
| `oidc.js` Microsoft issuer | Prefix check `iss.startsWith("https://login.microsoftonline.com/")` accepted the bare prefix with nothing after it — not a valid issuer | Added `iss.length > issuerPrefix.length` — tenantId segment must be present |

### False positives not changed

- Expiry check `payload.exp < now` — reviewer flagged `>` vs `>=`. At `exp = now`, `exp < now` is false → session accepted for that instant. `exp` is in whole seconds; `now` is `Math.floor(ms/1000)`. Sub-second grace is correct and expected OIDC behaviour. No change.
- Name fallback `email.split("@")[0]` — OIDC spec allows `name` to be absent. Fallback is intentional so the DB always has a non-empty name column.

### No remaining issues

PKCE generation, state cookie HMAC, token exchange (server-side only), claim validation (exp/iat/aud/iss/nonce), cookie flags (HttpOnly/Secure/SameSite) all correct.

nikshepkulli (approved):
Approving. The core OIDC properties are all correct:

- **PKCE**: `code_verifier` is 32 random bytes per-request (better than per-session), `code_challenge = base64url(SHA-256(verifier))`, matches RFC 7636 test vector in the test suite.
- **State**: 16 random bytes, HMAC-SHA256 signed into a 10-minute HttpOnly Secure SameSite=Lax cookie, callback re-verifies signature AND checks `payload.state === stateParam` AND `payload.provider === provider`. Belt-and-suspenders exp check in software beyond the cookie's Max-Age.
- **Nonce**: generated at start, embedded in state cookie, verified against ID token claim in callback.
- **Client secrets**: only ever in `env.*`, exchange happens server-side in `fetch(config.tokenEndpoint)`, no `client_secret` leaks to the browser.
- **ID token verification**: JWKS fetch → key selection by `kid` (or exactly-one signing key fallback) → RS256/ES256 signature verify → **alg allowlist** (`["RS256", "ES256"]`) closes the algorithm-confusion vector → `iss` + `aud` + `exp` + `iat` future-skew (300s) + `nonce` all checked. Clean.
- **Error path**: `loginFailed` returns a static 400 page with generic copy in the URL bar; the actual reason goes to `console.error` (Logpush) server-side. No info leak.

**Two follow-ups I'd file (both non-blocking, worth tracking):**

**1. `redirect_uri` derived from Host, not allowlisted.** `const origin = new URL(request.url).origin` trusts the Host header. In practice CF Pages restricts Host to registered hostnames and Google/MS's OAuth app registration rejects unknown `redirect_uri`s — two layers of defence — but an explicit `ALLOWED_ORIGINS` allowlist would harden this to defence in depth.

**2. No timeout on JWKS fetch.** `fetch(config.jwksUri, { cf: { cacheTtl: 3600 } })` will hang up to CF Workers' default 30s if Google or Microsoft's JWKS endpoint is slow. Wrap with `AbortController` + a 5s timeout so a JWKS outage doesn't stall every login. That's #161/#162 in the follow-up list — worth doing after this ships.

Also: `msTenant` pinning is deferred by design — you accept any tenant matching `https://login.microsoftonline.com/<tenantId>/v2.0`. Right call for a school (any Microsoft account fine); would be wrong for a B2B tenant-locked deployment. Ticket #160 tracks this if RCA ever wants to restrict.

Merge as part of the auth trio (after #154).

Issue #13, opened 2026-09-01
[Epic] UI rebuild-or-fix: the plan


Abhigna1975:
Tracking issue for the UI work on restcoderacademy, off the audit of this repo and the live deploy on 31 Aug 2026.

**The finding:** the site does not need rebuilding. The stack (React 19, Vite, MUI, componentised into atoms/molecules/organisms) is sound and the palette and typeface are RCA's own and work. What's wrong is plumbing and structure — a dead lead pipe, stale content, a 7MB page, and no page to send anyone to.

## Scope

**UI only.** Hosting (Cloudflare) and SEO are Nik's and are not in any of these tickets. Where a ticket touches something adjacent to his work — routes in #8, page structure — it delivers the UI for him to work against and stops there.

## The reference

`design/ui-template.html` and `design/tokens.css`, open for review as PR #14. Open the template in a browser. Every colour, size, radius and component state referenced below is defined there. Nothing in it is a new brand — it's the palette already in `src/styles/colors.css`, written down once.

## Order of work

**P0 — losing money right now, roughly a day together**

- [ ] #2 · Enquiry form silently loses every lead
- [ ] #3 · Batch dates hardcoded to 24-05-2025
- [ ] #4 · Landing page ships ~7MB of images

#2 is the one to do first. The backend has been unreachable and the form still says "Thanks, we will reach you soon" — so leads have been vanishing with no error anywhere. At a ₹35,000 fee, each one is a ₹35,000 enrolment nobody called back.

**P1 — what caps revenue, about a week**

- [ ] #5 · Navbar hamburger is a dead control at 600px and above
- [ ] #6 · Adopt design tokens: one typeface, four radii
- [ ] #7 · ButtonComponent has no hover, pressed, focus or loading state
- [ ] #8 · Per-course pages: wire react-router
- [ ] #9 · Hero: replace the mission paragraph, bring proof above the fold

#8 is the structural one. The site is a single page with no routes, so there is nowhere for a Java Full Stack visitor to land except a homepage where they scroll past the mission statement to find it.

**P2 — next**

- [ ] #10 · Placement logos hotlinked from third-party servers
- [ ] #11 · Responsive pass at 375 / 768 / 1024 / 1440
- [ ] #12 · Price and enrol CTA on the course card — blocked on Nik
- [ ] Enrolment and payment flow — design already drafted and open for review as PR #1 (`design/payment-gateway`)

## What's blocked and on whom

| | |
|---|---|
[...]

PR #157, opened 2026-09-07
design(social): rebuild the three recurring social templates in-repo (closes #151)

### LINKED ISSUE #151: [Design] Update social template watermarks from .com to .in — every historical post advertises a dead URL
## What's wrong

Every recurring branded social template still shows `www.restcoderacademy.com` as the URL watermark in the bottom bar / footer of the graphic. That domain returns nothing (400s / DNS null / dead). Verified on the live IG feed and in the LinkedIn Company Page admin view today (2026-09-07).

The consequence: **every historical placement announcement, festival greeting, batch launch, and Reel thumbnail RCA has ever published is advertising a URL that goes nowhere.** Any prospect who screenshots one of these — or reads it off IG on their phone and types the URL later — hits 404 and probably never comes back.

This is separate from the earlier fixes in the session:
- **Task #13** fixed the Google Business Profile "website" field that had been pointing at the dead .com
- **Task #15** fixed the www→apex Cloudflare redirect
- **PR #148** shipped the `/courses/fde` short-URL redirect

None of those touched the graphic templates themselves. The visual assets need to be edited by hand in the source file.

## Scope

Three recurring templates need the URL watermark corrected — from `www.restcoderacademy.com` (or any `.com` variant) → `restcoderacademy.in`:

1. **Placement Testimonial template** — blue gradient bg, student portrait cropped right, name + designation + Y.O.P + salary badge, quoted testimonial mid-panel, bottom bar with college / phone / URL. The URL is the .com problem.
2. **Festival Greeting template** — themed illustration bg, RCA "R" logo top-right, script/decorative headline, watermark URL in the corner. Same .com problem.
3. **Batch / Course Launch template** — navy blue with cyan accents, instructor portrait (Uday), feature grid, "Join Now" CTA. Watermark URL here too.

**Master file:** wherever the RCA brand kit lives (Figma / Canva / Adobe — pick the one that's the master source of the .fig / .pptx / .psd for these graphics).

## What's NOT in scope

- **Do NOT re-generate old posts.** Historical IG + LinkedIn posts stay as they are — going back to edit them will look like brand-panic and reset engagement metrics. Only fix the TEMPLATES so future posts use the correct URL.
- **Do NOT change any other design element** — colours, typography, layout, imagery, spacing. Only the URL text. Every other design decision is settled and working.
- **Do NOT update the .com watermark to include a protocol** (`https://…`). The current pattern is bare domain — keep that.
- **Do NOT add a QR code** or other new visual affordance to compensate for the URL confusion. Simple string swap only.

## Also worth catching while you're in the file

Two other small template corrections spotted during the audit that fit this same "brand-integrity string swap" bucket — do only if trivial in the same pass, otherwise skip:

- **The trailing `www.` prefix.** If the master template uses `www.restcoderacademy.com`, the corrected form should be `restcoderacademy.in` (drop the `www.` — see the Cloudflare www→apex 301 redirect from #15; keeping www would send prospects through an unnecessary hop). If it's already bare `.com`, just swap the TLD.
- **Phone numbers `9110424403` and `8073762257`** — verified during the IG audit as appearing on every placement template. Confirm both are Uday's and both are intended to be public. If either has changed since the template was made, correct it now.

## Done when

- [ ] Placement Testimonial master template shows `restcoderacademy.in` (no `www.`, no protocol)
- [ ] Festival Greeting master template shows `restcoderacademy.in`
- [ ] Batch / Course Launch master template shows `restcoderacademy.in`
- [ ] Master file saved back to whichever tool is the source of truth
- [ ] Optional: post-export a proof of one placement graphic and drop it in this ticket so we can eyeball it before the next real post
- [ ] The next real post that ships uses the corrected template

## Priority + effort

**P1.** Not a P0 code fix, but every day this stays untouched, more posts get published carrying a dead URL. Effort is a string swap in three template files — ~15-60 min once you're in the source file, depending on how the master is structured.

## Related

- Task #13, Task #15, PR #148 — earlier fixes for the same underlying dead-URL problem in different surfaces
- Content audit brief in `~/Desktop/RCA/restcoderacademy.in-audit/deliverables/28-content-audit-and-brief.md` — where this finding surfaced

Abhigna1975:
Closes #151.

## The problem

Every recurring social template watermarks `www.restcoderacademy.com`. That domain is dead, so every placement announcement, festival greeting and batch launch RCA has ever published advertises a URL that goes nowhere. Anyone who screenshots a post, or reads the URL off Instagram and types it later, hits nothing.

Task #13 fixed the Google Business Profile field, #15 the www→apex redirect, PR #148 the `/courses/fde` short URL. None of them touched the graphics, which is where most of the exposure actually is.

## Why this is code and not a Figma edit

The templates only ever existed inside a design tool. That is why the wrong URL survived three years and three adjacent fixes: it was invisible to the repo, unreviewable, and unfixable without opening the master file.

So the three templates are authored here instead:

| Path | What |
|---|---|
| `design/social-templates/brand.js` | the domain, the phones, the palette — **once** |
| `templates/*.js` | placement testimonial, festival greeting, batch launch |
| `tools/build.mjs` | → `out/*.svg`, `--png` proofs, `--check` for CI |
| `tests/social-templates.test.js` | fails if the dead URL ever comes back |

The domain is now one line. Templates call `watermark()` and never write it themselves, so a correction cannot land on two graphics and miss the third — which is exactly the failure this ticket is an instance of.

## Not a re-brand

Colours come from `design/tokens.css`, layouts are transcribed from the live graphics, nothing was redesigned. Per the ticket: historical posts untouched, no QR added, no protocol prefix, and `www.` dropped so nobody is sent through the #15 301.

## Templates take data, not fixed art

A missing salary collapses its badge instead of leaving an empty box. Long testimonials wrap and truncate rather than running off the canvas. Three features lay out without a gap where a fourth would be. Pasted text is escaped.

## Checks

- `npm test` — 88 passing, 13 new
- `npx eslint` — clean on every touched file
- PNG proofs of all three rendered and eyeballed at 1080×1080

PNG proofs are gitignored (they need a browser and only exist to eyeball a post). The SVGs are committed because they diff — the next change to a graphic shows up in review.

## One question for Uday

Both phone numbers on the placement template (`8073762257`, `9110424403`) came from the IG audit. Are both current, and are both meant to be public? They are one edit in `brand.js` if not.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ node_modules
 /.wrangler/
 /playwright/.cache/
 dist
+
+# Social-template PNG proofs — regenerate with tools/build.mjs --png (#151)
+design/social-templates/out/*.png
--- a/design/social-templates/README.md
+++ b/design/social-templates/README.md
@@ -1,0 +1,82 @@
+# Social templates
+
+The three recurring graphics RCA publishes, as source you can diff.
+
+| Template | File | Used for |
+|---|---|---|
+| Placement Testimonial | `templates/placement-testimonial.js` | every placement announcement |
+| Festival Greeting | `templates/festival-greeting.js` | Ugadi, Deepavali, Sankranti, … |
+| Batch / Course Launch | `templates/batch-launch.js` | new batch and course posts |
+
+## Why these live here
+
+Ticket #151: every historical placement announcement, festival greeting and
+batch launch carries `www.restcoderacademy.com` as its URL watermark. That
+domain is dead — the registration lapsed and the `.in` is the live site. Anyone
+who screenshots a post, or reads the URL off a phone and types it later, hits
+nothing.
+
+The templates only ever existed inside a design tool, so the wrong URL was
+invisible to the repo, unfixable without opening the master file, and printed
+identically onto three years of posts. Moving them here makes the domain one
+line in `brand.js` and the next correction a one-word diff that CI can check.
+
+This is **not** a re-brand. Colours come from `design/tokens.css`, the layouts
+are transcribed from the existing graphics, and nothing was redesigned. #151
+asked for a string swap; this is the machinery that keeps it swapped.
+
+## Change the URL, a phone number, the colours
+
+Edit `brand.js`. That is the only file with the domain in it — the templates
+call `watermark()` and never write it themselves, so a change cannot land on
+two graphics and miss the third.
+
+```bash
+node design/social-templates/tools/build.mjs           # rebuild out/*.svg
+node design/social-templates/tools/build.mjs --png     # + 1080×1080 PNG proofs
+node design/social-templates/tools/build.mjs --check   # fail if a dead URL is back
+```
+
+`out/*.svg` is committed because it diffs — a review shows exactly what moved
+on the graphic. `out/*.png` is not: it needs a browser, and it is only for
+eyeballing a proof before a post goes out.
+
+## Make a post
+
+Templates take data; the defaults are a worked example.
+
+```js
+import { render } from "./templates/placement-testimonial.js";
+
+const svg = render({
+  name: "Sakshi Rao",
+  designation: "Associate Engineer",
+  company: "Wipro",
+  yop: "2025",
+  salary: "₹5.2 LPA",
+  quote: "…",
+  college: "RV College of Engineering",
+});
+```
+
+Every field is optional — a missing salary collapses the badge rather than
+leaving an empty box, and long text wraps and truncates instead of running off
+the canvas. Drop the real photograph into the dashed `PHOTO` slot in any
+editor; the composition around it is already final.
+
+## Canvas
+
+1080×1080, Instagram's feed post. Stories and Reel covers crop from the
+centre, which is why nothing meaningful sits in the outer 96px.
+
+## What is deliberately not here
+
+- **Old posts are not re-generated.** Historical Instagram and LinkedIn posts
+  stay exactly as published. Going back to edit them reads as brand-panic and
+  resets the engagement on each one. Only the templates are fixed, so the next
+  post is right.
+- **No QR code, no `https://` prefix.** Both were considered in #151 and ruled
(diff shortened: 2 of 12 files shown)


nikshepkulli (approved):
LGTM. All the checks land where you'd want them to:

- **All three rendered SVGs (`out/*.svg`) show `restcoderacademy.in`, zero occurrences of `.com`, zero `www.`, zero protocol** — verified with grep against the branch head. The template source files (`templates/*.js`) never write the URL themselves; every graphic pulls it from `brand.js`, so the class of bug this ticket is an instance of cannot recur.
- The only surviving `.com` + `www.` references are in `brand.js`'s and `README.md`'s **comment prose explaining the fix**, not in anything the build emits. Correct.
- **Bare apex, no `www.`, no protocol** — matches the #151 spec exactly, and your comment on `SITE_URL` calls out why (`www.` triggers #15's 301, protocol wasn't in the source graphics).
- Phone numbers `8073762257` (PRIMARY) + `9110424403` (SECONDARY) preserved. **Answering your Uday question inline: both confirmed current + public in @nikshepkulli's WhatsApp reply this morning** — no edit needed.
- **Colours transcribed from `design/tokens.css`** into `COLOR`, not re-picked. Nothing else in scope was redesigned.

**On the scope creep** (ticket asked for a string swap; this PR is 12 files, +911 lines, moves the templates into the repo entirely): worth it. Your PR body captures why in one sentence — *"the wrong URL survived three years and three adjacent fixes because the templates only ever existed inside a design tool."* Fixing the string without fixing the invisibility is a fix that decays. This is the right level of scope for the failure mode.

**Nothing worth blocking on**, but two small notes for a follow-up if you want them:
- Committed SVGs must stay in sync with template JS. The `--check` mode in `tools/build.mjs` presumably fails CI on drift — worth confirming a CI job actually runs it.
- Consider adding a note to `README.md` on the workflow when a template changes (edit JS → run build → commit both). The pattern is right; documenting it stops the next contributor from editing the SVG directly.

Approving. Merge whenever.

PR #158, opened 2026-09-07
feat(placements): admin CRUD + D1-backed placement cards (closes #145)

### LINKED ISSUE #145: [Portal] Admin CRUD for placements — move site cards from hardcoded to D1
## Problem

Placement cards on the marketing site are hardcoded in `src/components/organism/placements/placement.js` (currently 4 records). The site placements are stale because the operational reality is: the team posts placements to Instagram in real time, but no one on the team is going through the "commit + PR + deploy" cycle to add them to the site. Instagram has many more placements than the site does, and every other channel (Justdial, LinkedIn) has fewer than either.

We considered two other approaches before landing here:

1. **IG Graph API sync via a Cloudflare Function** — rejected. Requires FB Business Manager setup, 2-4 week Meta app review, 60-day token refresh flow, ongoing token-expiry maintenance, and — most importantly — depends on the team consistently tagging every placement post with a distinguishing hashtag like `#RCAPlacement`. If the team isn't disciplined about that today (they aren't), a sync would pull nothing useful. Same underlying failure mode.
2. **Widget like Curator.io ($19/mo)** — rejected. Adds ~150KB JS bundle (regression on the CWV work we just shipped in #140 / #141 / #142 / #144), and the IG posts are visually-designed cards that clash with the current split-photo-and-logo site layout. Editorial control also lost.

This is the pragmatic path: give the team an admin form so a placement takes 5 minutes to publish to the site — the same effort as a well-crafted IG post — with full editorial control over which ones go on-site.

Fits into the portal admin surface already in scope for Phase 2 (#136 / #137).

### 1. D1 schema

New table `placements`, roughly:

- `id` (uuid, pk)
- `name` (text, required) — student name as shown publicly
- `designation` (text, required) — role at the company
- `company_name` (text, required)
- `company_logo_url` (text, required) — reference to uploaded asset or bundled logo path
- `photo_url` (text, required) — reference to uploaded student photo
- `description` (text, required) — student's own testimonial
- `background` (text, optional) — for the case-study on `/placements`
- `journey` (text, optional) — for the case-study on `/placements`
- `linkedin_url` (text, optional) — enriches Person JSON-LD via `sameAs`
- `course_slug` (text, optional) — link to `/courses/<slug>`
- `is_published` (boolean, default false)
- `display_order` (integer, default 0) — for manual ordering in the carousel
- `created_at`, `updated_at` (timestamps)

Migration seeds the current 4 hardcoded placements (Ashish, Sakshi, Sujith, Prajwala) as the initial rows so the site stays visually unchanged when the switch flips.

### 2. Admin screen — `/portal/admin/placements`

Behind the existing auth gate (session cookie from #114). Access limited to admin role.

- **List view** — table of all placements: name, company, published toggle, edit link, delete
- **Add / edit form** — fields listed above. Photo + company logo upload to Cloudflare R2 (or similar); URLs written to D1
- **Publish toggle** — an unpublished placement is invisible on the site until toggled on. Lets us curate rather than auto-publish
- **Display order** — either drag-to-reorder or an integer field, whichever is simpler

Nothing fancy — MUI table + form, same styling as the rest of the portal admin.

### 3. Public read API — `functions/api/placements/list.js`

`GET /api/placements/list` returns published placements as JSON, ordered by `display_order` then `created_at DESC`.

Cache-Control: `public, max-age=300` (5 min) — the site does not need real-time freshness for placements, and this dramatically reduces D1 read volume.

### 4. Site component updates

- `PlacementCard.jsx` / `Placements.jsx` — fetch from `/api/placements/list` on mount instead of importing from `placement.js`. Add a loading skeleton for the ~200ms fetch window (matching what the Reviews carousel already does)
- `PlacementsPage.jsx` — same. Person JSON-LD stays generated the same way, just from the fetched data
- Fallback: if the fetch fails (network error, function down), fall back to the hardcoded array so we never render an empty placements section

### 5. Delete `placement.js` last

Only after (1)-(4) are shipped and the seeded data is confirmed rendering on production. Delete `placement.js` in a follow-up PR — not this one — so the migration can be rolled back cleanly if anything goes wrong.

## Non-goals

- **Not doing IG sync in a second phase.** If this option works (and it should), a sync is unnecessary. If it fails because no one uses the admin form, a sync would fail for the same reason.
- **Not building a public "submit your placement" form.** Alumni-submitted content needs different moderation, and this is a curated marketing surface.
- **Not adding search / filter to the admin list.** The dataset is small (target: 30-40 records over the year); a plain table is fine.

## Done when

- [ ] `placements` table exists in D1 with the 4 current records seeded
- [ ] Admin at `/portal/admin/placements` can add, edit, publish/unpublish, and reorder — verified from a real admin session
- [ ] `GET /api/placements/list` returns the published set as JSON
- [ ] Homepage `Placements` carousel + `/placements` page both render from the API in production
- [ ] Adding a new placement via the admin form makes it appear on the live site within 5 min (accounting for the API's 5 min cache)
- [ ] The old `placement.js` is either kept as a fallback in the component OR removed in a separate follow-up PR
- [ ] Person JSON-LD on `/placements` still validates in Google Rich Results Test with fetched data

## Parent epics

Related to #41 (Student Learning Platform) and follows on from the admin work in #136 / #137.

Abhigna1975:
Closes #145.

## Why

Placement cards were hardcoded in `placement.js`, so publishing one meant a commit, a PR and a deploy. Nobody runs that cycle for a placement — which is why Instagram has many more than the site does, and the site's most persuasive section is permanently behind.

## What

| File | |
|---|---|
| `schema-placements.sql` | table + seed of the current records |
| `functions/api/placements/list.js` | `GET /api/placements/list`, published only, 5-min edge cache |
| `functions/admin/placements.js` | add / edit / publish toggle / display order / delete |
| `shared/placements.js` | one row↔record mapping and one validator, shared by both |
| `usePlacements.js` | fetches, falls back to the bundled array |

## Three things that differ from the ticket, on purpose

**The seed is five records, not four, and there's an `instagram_url` column.** Kota Akshay's record landed after #145 was written — embed-only, no photo/company/testimonial, because the IG embed carries all of it. Its comment in `placement.js` says it should migrate as-is, so it does. That's why `designation` and `company_name` are nullable, and why validation asks for *either* an embed URL *or* a designation-and-company: a row with neither renders an empty card.

**The four original records keep their images via `bundled:<key>`.** Their real URLs carry a Vite content hash that only exists after a build, so no literal URL could go in a seed. The client resolves the key against the same imports the fallback uses.

**The admin screen uses `requireAdminAuth`, not a portal session.** The ticket says "session cookie from #114" — that's the *student portal's* gate, and #114 hasn't landed. Every admin route on `main` uses the same Basic Auth, including the portal-users page from #112. Happy to switch it if you'd rather it wait for the portal.

## Not included

**R2 upload.** There's no R2 binding in `wrangler.toml`. Photo and logo take URLs, which is what the ticket's own schema specifies (`photo_url`, `company_logo_url`). Worth its own ticket if the team wants to upload straight from a phone.

## Failure behaviour

The list endpoint returns `{placements: []}` with a **200** when D1 errors, never a 500 — the site falls back to the bundled array on an empty list, and a 500 would turn "briefly stale" into "the section is broken".

New entries are **unpublished** until someone presses Publish, so nothing reaches the site unreviewed. The publish toggle is its own action, so taking a testimonial down (the urgent case) never depends on the rest of the record still validating.

## Verified

End to end against local D1 through `wrangler pages dev`:

- gate 401s unauthenticated reads **and** writes, and fails closed with no `ADMIN_PASSWORD`
- add → row exists, hidden from `/api/placements/list`
- publish → appears in the list
- invalid record (name only, or a `javascript:` URL) → no row written
- delete → gone
- `/placements` and the homepage carousel render from D1 with **zero broken images** and no console errors; JSON-LD is rebuilt from the fetched data (4 Review + 5 Person + ItemList), so structured data can't drift from what's on the page

22 new tests · 97 passing · build clean · lint clean on every new file.

### DIFF
--- a/functions/admin.js
+++ b/functions/admin.js
@@ -72,7 +72,7 @@ function page(notice, count, body) {
   .notice { padding:1rem 1.25rem; color:#b3261e; }
 </style></head>
 <body>
-  <header><h1>Enquiries</h1><span class="count">${count} total</span><a href="/admin/enrollments">Enrolments</a><a href="/admin/batches">Batches</a><a href="/admin/trainers">Trainers</a><a href="/admin/founder">Founder →</a></header>
+  <header><h1>Enquiries</h1><span class="count">${count} total</span><a href="/admin/enrollments">Enrolments</a><a href="/admin/batches">Batches</a><a href="/admin/trainers">Trainers</a><a href="/admin/placements">Placements</a><a href="/admin/founder">Founder →</a></header>
   ${notice ? `<div class="notice">${escapeHtml(notice)}</div>` : ""}
   <div class="wrap">
     <table>
--- a/functions/admin/batches.js
+++ b/functions/admin/batches.js
@@ -127,7 +127,7 @@ function page(rows, error, notice) {
   .hint{font-size:.75rem;color:var(--muted);margin:.3rem 0 0}
 </style></head>
 <body>
-  <header><h1>Batches</h1><a href="/admin/enrollments">Enrolments</a><a href="/admin/trainers">Trainers</a><a href="/admin/batches">Batches</a><a href="/admin/founder">Founder</a><a href="/admin">Enquiries →</a></header>
+  <header><h1>Batches</h1><a href="/admin/enrollments">Enrolments</a><a href="/admin/trainers">Trainers</a><a href="/admin/batches">Batches</a><a href="/admin/placements">Placements</a><a href="/admin/founder">Founder</a><a href="/admin">Enquiries →</a></header>
   <div class="wrap">
     ${error ? `<div class="banner err">${escapeHtml(error)}</div>` : ""}
     ${notice ? `<div class="banner ok">${escapeHtml(notice)}</div>` : ""}
--- a/functions/admin/enrollments.js
+++ b/functions/admin/enrollments.js
@@ -79,7 +79,7 @@ function page(notice, count, body) {
   .amt { font-weight:600; white-space:nowrap; }
 </style></head>
 <body>
-  <header><h1>Enrolments</h1><span class="count">${count} total</span><a href="/admin/trainers">Trainers</a><a href="/admin/batches">Batches</a><a href="/admin/founder">Founder</a><a href="/admin">Enquiries →</a></header>
+  <header><h1>Enrolments</h1><span class="count">${count} total</span><a href="/admin/trainers">Trainers</a><a href="/admin/batches">Batches</a><a href="/admin/placements">Placements</a><a href="/admin/founder">Founder</a><a href="/admin">Enquiries →</a></header>
   ${notice ? `<div class="notice">${escapeHtml(notice)}</div>` : ""}
   <div class="wrap">
     <table>
--- a/functions/admin/founder.js
+++ b/functions/admin/founder.js
@@ -97,7 +97,7 @@ function page(r, error, notice) {
   .hint{font-size:.8rem;color:var(--muted);margin:.3rem 0 1rem;line-height:1.5}
 </style></head>
 <body>
-  <header><h1>Founder page</h1><a href="/admin/enrollments">Enrolments</a><a href="/admin/trainers">Trainers</a><a href="/admin/batches">Batches</a><a href="/admin/founder">Founder</a><a href="/admin">Enquiries →</a></header>
+  <header><h1>Founder page</h1><a href="/admin/enrollments">Enrolments</a><a href="/admin/trainers">Trainers</a><a href="/admin/batches">Batches</a><a href="/admin/placements">Placements</a><a href="/admin/founder">Founder</a><a href="/admin">Enquiries →</a></header>
   <div class="wrap">
     ${error ? `<div class="banner err">${escapeHtml(error)}</div>` : ""}
     ${notice ? `<div class="banner ok">${escapeHtml(notice)}</div>` : ""}
--- a/functions/admin/placements.js
+++ b/functions/admin/placements.js
@@ -1,0 +1,234 @@
+// /admin/placements — publish a placement to the site without a deploy (#145).
+//
+// Same Basic Auth as the rest of the admin panel (ADMIN_PASSWORD, fails closed
+// if unset). GET renders every placement with an edit form each plus an add
+// form; POST handles add / update / publish toggle / delete, writes D1, and
+// redirects back.
+//
+// Why this exists: the team announces placements on Instagram the day they
+// happen, and nobody runs a commit → PR → deploy cycle to mirror them onto the
+// site — so the site's most persuasive section is permanently behind Instagram.
+// Publishing here should cost about what posting to Instagram costs.
+import { escapeHtml, requireAdminAuth } from "../../shared/serverUtil.js";
+import { EDITABLE_FIELDS, inputToColumns, validatePlacement } from "../../shared/placements.js";
+
+const SELECT =
+  "SELECT id, name, designation, company_name, company_logo_url, photo_url, " +
+  "description, background, journey, linkedin_url, course_slug, instagram_url, " +
+  "is_published, display_order FROM placements " +
+  "ORDER BY display_order ASC, created_at DESC";
+
+function html(body, status = 200) {
+  return new Response(body, {
+    status,
+    headers: { "Content-Type": "text/html; charset=utf-8", "cache-control": "no-store" },
+  });
+}
+
+function redirect(request, param, msg) {
+  const url = new URL(request.url);
+  url.search = "";
+  url.searchParams.set(param, msg);
+  return Response.redirect(url.toString(), 303);
+}
+
+// ---------------------------------------------------------------------------
+// GET
+// ---------------------------------------------------------------------------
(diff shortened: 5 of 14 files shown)


nikshepkulli (approved):
LGTM. Every concern the ticket + my earlier PoC (#152) raised is addressed correctly; the failure-mode philosophy is exactly right for the section that does the persuading. Approving.

## What I verified

- **Seed has Kota Akshay Rathna Kumar with `https://www.instagram.com/p/Dc5o-Uatcxz/`** exactly as #152 shipped. Column name is `instagram_url` — matches.
- **All four original placements** (Ashish / Sakshi / Sujith / Prajwala) seeded with their `description` bodies **verbatim** from `placement.js`. Zero drift.
- **`InstagramEmbed.jsx` preserved** and still imported by both `PlacementCard.jsx` and `PlacementsPage.jsx`. My #152 work absorbed cleanly.
- **No file conflicts with #152** — `placement.js` retained as the fallback, extended with `BUNDLED_ASSETS` + `resolveAsset()`; `InstagramEmbed.jsx` untouched.
- **Parameterised queries throughout** (`.bind()` in every prepared statement) — no injection surface.
- **Input bounded** at 4000 chars per field in `inputToColumns` — no unbounded blob writes.
- **HTML escaping** consistent in the admin renderer (`escapeHtml` on every dynamic insert including error/notice banners).
- **URL validation** rejects `javascript:` (only `https?://` or `/` or `bundled:` passes); IG URL host-locked to `instagram.com`.
- **Delete has `confirm()` guard**, publish-toggle is its own action so takedown never requires re-validating other fields.
- **5s AbortController timeout** on the fetch, `finally` clears the timer.
- **`bundled:<key>` for the four seeded photos** is a genuinely elegant solution to the Vite content-hash problem — anything added through the admin form takes a plain URL, so the special case never propagates.
- **JSON-LD rebuilt from fetched data** in `PlacementsPage.jsx` — Person + Review + ItemList can't drift from what's rendered.
- Public list endpoint returns `{placements: []}` with **200** on any error, so the fallback array kicks in. A 500 would render the section broken. Right call.
- 22 tests including the negative cases (IG URL that isn't instagram.com, `javascript:` URLs, name-only rows, corrupt payloads).

## The three deliberate deviations you flagged

All three are the right call:

1. **Fifth seed record + nullable `designation`/`company_*`** — Kota's IG-embed-only card. Matches how #152 shipped him.
2. **`bundled:<key>` for original images** — Vite hash constraint. Correct.
3. **`requireAdminAuth` (Basic Auth) instead of #114 portal session** — #114 hasn't landed and every other `/admin/*` route is Basic Auth. Consistency wins.

## Follow-up worth filing (not blocking)

**YouTube video / LinkedIn post URL columns not included.** My 2026-09-07 comment on #145 spec'd those alongside `instagram_url`. Row shape currently ships without them; adding either later needs a migration. Worth its own ticket once we know whether the team actually produces alumni videos.

Merge whenever.

nikshepkulli:
Merged. Handing off two follow-up steps to you, @Abhigna1975 — both need CF access I don't have and shouldn't have:

**1. Apply the D1 migration.**
The public list endpoint returns 200 with an empty array on any DB error and the frontend falls through to the bundled `placement.js` fallback, so the site keeps rendering the current 4 placements until this runs. But `/admin/placements` will 500 on first attempt without the table. Command is something like:

```
wrangler d1 execute <binding-name> --file=schema-placements.sql --remote
```

You know the exact binding name from your dev setup — replace `<binding-name>` with what's in `wrangler.toml`.

**2. Set the admin Basic Auth credentials on CF Pages.**
The env var names live at the top of `functions/admin/placements.js` (matching the pattern the other `/admin/*` routes use — you'd know better than me which pattern that is, `ADMIN_USER`/`ADMIN_PASS` or similar). Set them in the Cloudflare Pages dashboard for `restcoderacademy.in` under Settings → Environment variables. Production environment. Once saved, re-deploy is automatic on the next push (or trigger one manually).

**Ping me when done** and I'll walk Uday through his first placement add. Meanwhile I'll message him separately to hold off publishing through the panel until you give the green light — placements he adds via IG in the interim stay valid and get picked up on the next admin session.

nikshepkulli:
Update: Cloudflare invite sent to `abhigna@kulli.co` — Administrator role, entire account scope (covers Pages env vars + D1 execute). Check your inbox for the CF invite email (should arrive in a few minutes).

After you accept and set up `wrangler login`, the two commands you'll need:

1. **Apply the D1 migration:**
   ```
   wrangler d1 execute <binding-name> --file=schema-placements.sql --remote
   ```
   Replace `<binding-name>` with what's in `wrangler.toml`.

2. **Set admin Basic Auth credentials in the CF Pages dashboard:**
   - Go to Pages → `restcoderacademy-in` project → Settings → Environment variables → Production
   - Set the env var names that `functions/admin/placements.js` reads (I didn't grep for the exact names — you'll know them from your dev setup)
   - Save; a re-deploy triggers automatically

Ping me here or on WhatsApp when both are done and I'll walk Uday through his first placement add.

Issue #150, opened 2026-09-06
[P1] Java & Python course pages not prerendered — serving 2.3KB SPA shell instead of full HTML


nikshepkulli:
## Finding

Discovered while verifying PR #148 live on production 2026-09-06:

\`\`\`
curl -s https://restcoderacademy.in/courses/full-stack-java     | wc -c    # 2340 bytes
curl -s https://restcoderacademy.in/courses/full-stack-python   | wc -c    # 2340 bytes
curl -s https://restcoderacademy.in/courses/mern-stack          | wc -c    # 45988 bytes
curl -s https://restcoderacademy.in/courses/forward-deployed-engineering | wc -c # 49055 bytes
\`\`\`

The Java and Python course pages return only the SPA index shell. MERN and FDE prerender correctly. Google, Bing, and every AI crawler see empty pages for the two most SEO-valuable slugs.

## Why this matters

These are the primary SEO targets:
- \`full-stack-java\` — matches "full stack java course bengaluru" / "java coding classes"
- \`full-stack-python\` — matches "full stack python course bengaluru" / "python coding classes"

Every piece of SEO work landed this session (per-course schema in #5, keyword-aligned H1 in #20, blog depth linking back to these pages in #7 / #21, LocalBusiness schema in #20, AggregateRating in #25) is **negated** for a crawler that sees only the SPA shell — because none of it is actually in the served HTML for those two routes.

Not introduced by any recent PR; predates #148. Discovered because #148's CTA text was missing from Java page HTML when spot-checking cross-course behavior.

## Where to start investigation

- \`scripts/prerender.mjs\` — the Playwright per-route prerender script; it iterates over routes and writes static HTML to \`dist/courses/<slug>.html\` for CF Pages to serve as the exact-static-match
- Compare what routes get prerendered vs \`courses/courses.js\` slug list. If \`full-stack-java\` and \`full-stack-python\` aren't in the prerender iteration, the fix is one line in the route list. If they ARE in the iteration but silently failing during render (Playwright timeout, throwing script), that's a different fix
- Check \`dist/courses/\` after a fresh \`npm run build\` — are the \`.html\` files actually being written for all four slugs?
[...]

PR #164, opened 2026-09-08
fix(seo): per-page Open Graph + Twitter cards, and a real card image (closes #73)

### LINKED ISSUE #73: [P1] Per-page titles, meta descriptions, canonicals + Open Graph
One static <title> today, no description, no social cards. Add per-route head management (react-helmet-async or the SSG head API): keyworded title, written meta description, canonical URL, and Open Graph/Twitter tags (image, title, description) so shared links look professional.
**Done when:** Each page has a unique title+description; a WhatsApp/LinkedIn share shows a proper card.
—
Part of #68 · [Growth & SEO audit](https://claude.ai/code/artifact/edfa9a4b-c5c1-4b83-8aed-b82c1305bf3f)

Abhigna1975:
Closes #73.

## The bug

**All 17 routes served the homepage's social card.** `index.html` ships one static set of og/twitter tags with the homepage's values, and nothing in `src/` ever overrode them:

```
/                          og:url = https://restcoderacademy.in/
/courses/java-full-stack   og:url = https://restcoderacademy.in/
/blog/java-vs-python...    og:url = https://restcoderacademy.in/
```

`og:title`, `og:description`, `twitter:title` and `twitter:description` were identical on every page too.

So sharing any of the 16 non-homepage URLs on WhatsApp or LinkedIn showed the homepage title, the homepage description, and a link that resolved to the front door. For a business that distributes through WhatsApp and LinkedIn, every shared course page and blog post pointed back at the entrance.

## What was already done

Half of #73 had landed: per-route `<title>`, `<meta name="description">` and `<link rel="canonical">` are correct across all 17 routes. Only the og/twitter half was missing, which is why the ticket's done-when was not met.

## The fix

- **`SocialMeta`** renders the per-route tags, following the pattern the pages already use for title and canonical. No head library introduced.
- **`prerender.mjs`** now `keepLast`s each of them. Without that, both the static and per-route tag survive and a crawler reading the first wins gets the homepage card again.
- **Each page's title moved into a const** shared by `<title>` and the card, so the two cannot drift. Same failure in miniature; one line per page to make impossible.
- **A real card image.** `og:image` was `favicon.png` at **624×681** — with `summary_large_image` a near-square image is letterboxed or centre-cropped, so a shared link looked broken rather than designed. `public/og-card.png` is 1200×630, generated by `scripts/og-card.mjs` from `design/tokens.css` so it tracks the brand, drawn with the mark alone (the wordmark is unreadable at thumbnail size), and built from flat shapes rather than a gradient so it survives WhatsApp's recompression.

## Verified across all 17 prerendered routes

- per-page `og:url` and `og:title`
- `og:url` matches the canonical exactly on every page
- `og:type: article` on blog posts, `website` everywhere else
- exactly one of each tag, no duplicates
- `/og-card.png` serves 200

97 tests passing, build clean, lint clean on every new file.

## On the test

It asserts the invariant at **source level**, not over built HTML. The failure mode here is not a malformed tag, it is someone adding a page and forgetting — which is exactly how this happened. A test over `dist/` would also need a build to run first.


nikshepkulli (approved):
LGTM. Clean root-cause fix on what it says it fixes; the `keepLast` in `prerender.mjs` is the load-bearing insight and it's implemented correctly.

## What I verified

- **`SocialMeta` follows the existing pattern** — no head library, just React 19's hoisting of title/meta into `<head>`, same way every page already sets its own `<title>` + description + canonical.
- **The `keepLast` in `scripts/prerender.mjs`** is the crucial piece I was watching for. Without it, both the static homepage tag from `index.html` AND the per-route one from `SocialMeta` would survive in the built HTML, and crawlers that take the first win would get the homepage card again — the exact bug being fixed. Correctly implemented for all 8 og tags + 4 twitter tags.
- **`og:type = "article"` propagates to `BlogPost.jsx` only** — verified in the test suite. Everywhere else is `website`.
- **1200×630 card is programmatically generated** from `design/tokens.css` via `scripts/og-card.mjs` — reproducible, tracks the brand, drawn with the mark alone (wordmark is unreadable at thumbnail size). Right call.
- **Both `og:image:width` and `og:image:height` declared** — LinkedIn and X need both to render full-width; without them they fall back to a small thumbnail.
- **Twitter cards fixed alongside OG** — both platforms covered, not just Facebook.
- **Bundled at `public/og-card.png`** — not hotlinked, matches the no-third-party-hosting convention (#10).
- **Tests are source-level assertions** — the whole point being that "someone adds a page and forgets" is exactly how this bug happened. A build-artifact test would need the build to run first; this catches the same class of mistake in the file where it's made. Includes the anti-pattern check (`no page hardcodes its own og: tags`) that stops the invariant from drifting later.

## Cross-check: PR #150 not affected

Live-verified before writing — `/courses/full-stack-java` and `/courses/full-stack-python` still return 2340-byte SPA shells on prod (MERN + FDE prerender fully at ~46–48KB). The `keepLast` edit in `prerender.mjs` only adds meta-tag persistence rules; it does NOT change which routes get iterated for prerender. So #150 is untouched — still an open problem, still needs its own fix.

Worth calling out because the tests in this PR assert `SocialMeta` is imported in `CourseDetail.jsx`, but for Java and Python that component never actually renders on the built HTML (because those two routes aren't prerendered at all). So on-prod after this merges, Java and Python course pages will still ship the homepage card via `index.html` until #150 is closed.

Not a blocker — this PR is doing what it should. Just a reminder that #150 is where the completion of #73 actually lives for those two slugs.

Merge whenever.

PR #166, opened 2026-09-08
fix(enroll): bind the recorded course to the payment that was actually made (closes #165)

### LINKED ISSUE #165: [P0] Enrolment records the course the browser claims, not the one that was paid for
Found while sweeping the payment path. **Reproduced against the real handler**, not theoretical.

## What is wrong

`/api/enroll/verify` verifies the Razorpay signature correctly, and the signature is genuine. But Razorpay's signature is an HMAC over **`order_id|payment_id` only** — it does not cover the course.

`verify.js` then takes the course from the request body:

```js
const courseId = String(b.course || "").trim();
const amount = priceForCourse(courseId);
```

and writes that course and that price into `enrollments` with `status='paid'`. Nothing ever compares the claimed course against the order that was actually paid.

## Why it matters now

There are four payable courses at two prices (`shared/enroll.js`):

| id | price |
|---|---|
| `fde` | ₹50,000 |
| `java-fs`, `python-fs`, `mern` | ₹35,000 |

So a real payment for a ₹35,000 course, re-submitted with `course: "fde"`, is recorded as a paid ₹50,000 FDE enrolment. The signature validates because it never covered the course. **₹15,000 per instance**, and `/admin/enrollments` shows them as fully paid.

This was harmless when FDE was the only payable course, which is what the comment at the top of `order.js` still assumes ("currently FDE"). Adding the other three made it live.

## Confirmed

Ran the real `onRequestPost` with a genuine order id, payment id and valid signature, and a swapped course:

```
paid for java-fs : 3500000 paise
recorded as      : fde 5000000 paise
status           : paid
```

## The fix, and it is cheap

`verify.js` **already fetches the payment from Razorpay** for the email and phone. That response carries `order_id`, `amount` and `status`. So:

- reject unless `payment.order_id` equals the submitted `razorpay_order_id`
- reject unless `payment.amount` equals `priceForCourse(courseId)`
- require `payment.status === "captured"`
- record `payment.amount`, the amount actually captured, rather than the price of the claimed course

The one thing to be careful about: that lookup is currently best-effort and its failure is swallowed, which is right for the email and phone. It cannot be best-effort for the amount check. If Razorpay cannot be reached, the payment should still be acknowledged to the student (they have paid) but the row must not assert a course and price we have not confirmed.

## Not affected

`order.js` is correct: the amount is looked up server-side from the course id and never read from the body, so nobody can pay less than the sticker price. This is only about what gets *recorded* afterwards.

Taking this now.

Abhigna1975:
Closes #165. **P0 — money.**

## The bug

Razorpay's signature is an HMAC over `order_id|payment_id` only. It does not cover the course. `verify.js` took the course from the request body and wrote it, with that course's price, as a paid enrolment. Nothing compared the claim against the order that was actually paid.

With four payable courses at two prices, a genuine ₹35,000 payment re-submitted naming the ₹50,000 course was recorded as a paid FDE enrolment, with a signature that validated.

Reproduced against the real handler before fixing:

```
paid for java-fs : 3500000 paise
recorded as      : fde 5000000 paise
status           : paid
```

**₹15,000 a time**, and `/admin/enrollments` showed them as fully paid. It was harmless while FDE was the only payable course — which is what the comment in `order.js` still assumes — and adding the other three made it live.

## The fix

`verify.js` already fetched the payment for the student's email and phone, so the authority was on hand and unused. It now also requires the payment to:

- belong to the order that was named
- be `captured`, not merely authorized
- match that course's price

and records **the amount actually captured**, not the price of the claimed course.

## The part worth reading carefully

Those two lookups have different failure rules:

- **Contact details** stay best-effort. A student who has paid must never be failed because a lookup hiccupped.
- **What they paid for** cannot be best-effort, or the check is worthless.

So when Razorpay is unreachable the payment is still acknowledged to the student, but the row records **no course and no amount** and is marked `needs_review` rather than asserting something unverified.

That needed handling at the other end too: `/admin/enrollments` rendered anything that was not `"paid"` as `"registered"`, so an unconfirmed payment would have shown a student who **has** paid as one who has not. It gets its own badge and shows the payment id to reconcile against Razorpay.

## Not affected

`order.js` is untouched and was already correct: the amount charged is looked up server-side from the course id and never read from the body. Nobody could ever pay *less* than the sticker price. This is only about what gets recorded afterwards.

## On the two tests I changed

Both asserted the old contract. One of them carried the comment `// server-decided, not from the body` while asserting the price of the body's claimed course, which is precisely the bug. The DB stub also keyed off a literal `'paid'` in the SQL that is now a bound parameter.

Nine new tests, including the original attack. 84 passing, build clean.

### DIFF
--- a/functions/admin/enrollments.js
+++ b/functions/admin/enrollments.js
@@ -32,17 +32,28 @@ function rupees(paise) {
 
 function rowHtml(r) {
   const paid = r.status === "paid";
+  // A payment we could not confirm with Razorpay (#165). The money is real, but
+  // we could not check which course it was for, so the row deliberately holds
+  // no course and no amount. Rendering it as "registered" like everything else
+  // that is not "paid" would show a student who HAS paid as one who has not —
+  // it needs its own badge and someone to reconcile it against the Razorpay
+  // dashboard using the payment id.
+  const review = r.status === "needs_review";
   const badge = paid
     ? `<span class="badge paid">paid</span>`
-    : `<span class="badge reg">registered</span>`;
+    : review
+      ? `<span class="badge review">needs review</span>`
+      : `<span class="badge reg">registered</span>`;
   return (
     "<tr>" +
     `<td>${escapeHtml(r.fullname)}</td>` +
     `<td><a href="tel:${escapeHtml(r.mobile)}">${escapeHtml(r.mobile)}</a></td>` +
     `<td>${r.email ? `<a href="mailto:${escapeHtml(r.email)}">${escapeHtml(r.email)}</a>` : "—"}</td>` +
-    `<td>${escapeHtml(r.course_name || r.course)}</td>` +
+    `<td>${escapeHtml(r.course_name || r.course) || "<em>unconfirmed</em>"}</td>` +
     `<td>${escapeHtml(r.batch) || "—"}</td>` +
-    `<td>${badge}${paid ? ` <span class="amt">${rupees(r.amount)}</span>` : ""}</td>` +
+    `<td>${badge}${typeof r.amount === "number" ? ` <span class="amt">${rupees(r.amount)}</span>` : ""}` +
+      (review ? ` <span class="pid">${escapeHtml(r.razorpay_payment_id)}</span>` : "") +
+      "</td>" +
     `<td>${escapeHtml(r.referral) || "—"}</td>` +
     `<td class="when">${escapeHtml(r.created_at)}</td>` +
     "</tr>"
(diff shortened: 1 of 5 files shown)


nikshepkulli (approved):
Approving. This is the right shape of fix for the vulnerability described.

**Root cause is real:** Razorpay's signature is an HMAC over `order_id|payment_id`, not the payload. Trusting the client-supplied course + price in `verify.js` was the gap.

**The fix is narrow:** `verify.js` now requires the fetched payment to (a) belong to the order it was submitted with, (b) be `captured` not authorized, (c) match the expected price for the course claimed — and records the actual captured amount. Nine tests including the reproduction. Nothing else changed.

**Ships correctly split by criticality:** contact details stay best-effort so a real paying student never gets failed by a lookup hiccup; course-and-price is authoritative because "best effort" there would defeat the check. That distinction is the right call.

Merging immediately — this is live in prod.

**Follow-up you (@nikshepkulli) should run manually:** as Abhigna noted, existing rows in `/admin/enrollments` should be reconciled against the Razorpay dashboard, since anything already recorded via this bug would look legitimate. Filing a follow-up ticket for that reconciliation would be worth it.

PR #172, opened 2026-09-08
refactor(courses): move the course card 16 off-token colours onto tokens (closes #170)

### LINKED ISSUE #170: [UI] Course card: move the 16 off-token colours in CourseCard.css onto tokens
Part of #13. The course cards are the benchmark for the rest of the site, so their own stylesheet should be the cleanest — right now it is the one carrying the most off-token colour.

## Context

The visual half of "lift Java/Python/MERN toward the FDE card's polish" is **already done**: #56 made `CoursesCard.jsx` a single uniform card for every course, with `flagship` reduced to a data-only prop that picks the syllabus arrays and the trainer credibility line, with no visual difference. Navy header, price slot, equal heights and the same CTA pair on all four cards.

What did not follow is token adoption. `CourseCard.css` has **16 bare colour values** with no token behind them:

| Line | Value | Note |
|---|---|---|
| 9, 134 | `#e6e9f2` | card + syllabus border — `--rca-border` is `#DDE3E7` |
| 21, 53, 198 | `#fff` | should be `--rca-ink-invert` |
| 36 | `#aeb6e0` | muted copy on navy — `--rca-ink-invert-soft` (`#C0C4DA`) exists for exactly this |
| 44 | `rgba(255,255,255,0.13)` | divider on navy — no token |
| 63 | `#8e96c8` | second muted-on-navy value, different from line 36 |
| 66, 76 | `#ffb74d` | a lighter orange than `--rca-orange` (`#FF9800`); also the focus ring colour |
| 116 | `rgba(255,152,0,0.14)` | orange tint, no token |
| 117 | `#8a5200` | dark orange text, no token |
| 133 | `#f5f7fb` | subtle surface — `--rca-surface-subtle` is `#F8FAFB` |
| 154, 157 | `#45545d` | `--rca-ink-soft` is exactly this |
| 185 | `#333` | `--rca-ink` is exactly this |

Four of these (`#45545d`, `#333`, `#f5f7fb`, `#e6e9f2`) are a token's value typed out literally — straight swaps. The rest are genuinely new values invented outside `tokens.css`, which is the thing the epic is trying to stop: two different muted-on-navy greys and an orange that is not the brand orange.

Note the `var(--token, #fallback)` pattern already used elsewhere in the file is fine and should stay; this is only about the bare values.

## What "done" looks like

- [ ] The four exact matches replaced with their tokens
- [ ] `#fff` → `--rca-ink-invert`; `#aeb6e0` → `--rca-ink-invert-soft`
- [ ] The two muted-on-navy greys reconciled to one value
- [ ] A decision on `#ffb74d`: either use `--rca-orange`, or add a named token for the lighter tint if the contrast on navy needs it. Same for the `rgba(255,152,0,0.14)` / `#8a5200` pair — if that badge treatment stays, it earns tokens.
- [ ] Any new token added to `design/tokens.css`, not to the component
- [ ] The divider on navy gets a token or a documented reason to stay inline
- [ ] No visual change intended — this is a substitution pass. Compare before/after at 375 / 768 / 1440.
- [ ] Focus-visible ring still meets contrast on the navy header

Design authority is `design/tokens.css` + `design/ui-template.html`.

Abhigna1975:
Part of #13. Closes #170.

The course cards are the benchmark the rest of the site is being brought up to, so their own stylesheet should be the cleanest. It was carrying the most colour that `design/tokens.css` knows nothing about — 16 bare values.

## Straight swaps

The file had a token's value typed out literally: `#45545d` → `--rca-ink-soft`, `#333` → `--rca-ink`, `#fff` → `--rca-ink-invert`.

## Snapped to the nearest token

A shift of a few units each: `#e6e9f2` → `--rca-border`, `#f5f7fb` → `--rca-surface-subtle`, and the `rgba(255,255,255,0.13)` divider → `--rca-tint-on-dark` (0.15).

## Deleted rather than converted

`.cc-tag` and `.cc-tag--muted` — no JSX references either class. They were the only reason `rgba(255,152,0,0.14)` and `#8a5200` were in the file.

## Two decisions worth a look

This is **not** the zero-diff pass the ticket assumed, and I would rather say so than bury it.

**The two muted greys on the navy header collapse to one token.** `.cc-sub` used `#aeb6e0`, the trainer's title line used `#8e96c8` — two invented values doing the same job at different strengths. Both are now `--rca-ink-invert-soft`. That is visible, and it is the change worth having:

| | before | after |
|---|---|---|
| `.cc-sub` (16px) | 9.25:1 | **10.63:1** |
| trainer title (11.5px, smallest text on the card) | 6.43:1 | **10.63:1** |

**The amber accent becomes the brand orange.** `#ffb74d` was a lighter orange than `--rca-orange`, used for the founder link and the focus ring. On navy it measured 10.61:1; `--rca-orange` measures 8.52:1 — lower, but far above the 4.5:1 AA wants for text and the 3:1 for a focus indicator, and it removes a second orange from the palette.

Happy to keep `#ffb74d` as a named token instead if you would rather hold that exact tone — it is a one-line change.

## Verification

Screenshot diff of the flagship card at 1440px: **6.16% of pixels differ, all of it the muted copy on the navy header, max channel delta 50.** No layout, size or spacing moved. Build clean, 177/177 tests pass.

The pre-existing `var(--token, #fallback)` pattern elsewhere in the file is left alone — the fallbacks are harmless and churning them is not this ticket.

### DIFF
--- a/src/components/organism/courses/CourseCard.css
+++ b/src/components/organism/courses/CourseCard.css
@@ -6,7 +6,7 @@
   display: flex;
   flex-direction: column;
   background: var(--rca-surface, #fff);
-  border: 1px solid #e6e9f2;
+  border: 1px solid var(--rca-border);
   border-radius: var(--rca-radius-lg, 12px);
   overflow: hidden;
   box-shadow: var(--rca-shadow-card, 0 1px 3px rgba(3, 8, 76, 0.08));
@@ -18,7 +18,7 @@
 /* --- Navy header --- */
 .cc-head {
   background: var(--rca-navy, #03084c);
-  color: #fff;
+  color: var(--rca-ink-invert);
   padding: 20px 24px 18px;
 }
 .cc-title {
@@ -33,7 +33,7 @@
      it takes the body size here too, rather than 13px (#11). */
   font-size: var(--rca-fs-body);
   line-height: 1.5;
-  color: #aeb6e0;
+  color: var(--rca-ink-invert-soft);
 }
 .cc-trainer {
   display: flex;
@@ -41,7 +41,7 @@
   gap: 11px;
   margin-top: 15px;
   padding-top: 14px;
-  border-top: 1px solid rgba(255, 255, 255, 0.13);
+  border-top: 1px solid var(--rca-tint-on-dark);
 }
 .cc-avatar {
   width: 36px;
@@ -50,7 +50,7 @@
   border-radius: 999px;
   overflow: hidden;
   background: var(--rca-blue, #146389);
-  color: #fff;
+  color: var(--rca-ink-invert);
   display: flex;
   align-items: center;
   justify-content: center;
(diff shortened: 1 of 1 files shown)


nikshepkulli (approved):
LGTM. Sixteen off-token colors → tokens, with `.cc-tag` / `.cc-tag--muted` removed on the way (assuming unused — CI green + no template refs). E2E passing means no `--rca-*` token was referenced that doesn't exist in tokens.css.

Approving. Ready to merge.

Issue #163, opened 2026-09-07
[Portal] Sign-in cannot work inside the Android app — the session never reaches the WebView


Abhigna1975:
Found while finishing the portal UI (#110/#111). This is a **Phase 1 (#42) blocker for the app specifically** — the web portal is unaffected.

#42's acceptance line is *"install the app, tap Sign in with Google/Microsoft, land on the student home."* As designed today that cannot happen, for two reasons.

## 1. Nothing is wired to the deep link, on either end

#109 registered the `rca://` scheme in `AndroidManifest.xml` with `launchMode="singleTask"`, and stopped there — correctly, it says so itself.

Since then:

- **No server route redirects to `rca://`.** Neither the callback in #155 nor the one that was in #127 mentions the scheme. Both end by redirecting to a normal same-origin path.
- **No JavaScript listens for it.** `@capacitor/app` is not in `package.json` — `npx cap sync` reports exactly one plugin, `@capacitor/splash-screen` — so there is no `appUrlOpen` listener and nothing could receive the URL even if it were sent.

So the scheme is registered and connected to nothing at both ends.

## 2. The harder half: a browser cookie is not a WebView cookie

This is the part worth designing before anyone writes code.

Google blocks OAuth in embedded WebViews (`disallowed_useragent`), so the app cannot run the consent flow inside its own WebView. It has to open a Custom Tab / system browser. But then:

- The callback runs in the **system browser**, and `Set-Cookie: __session=…` lands in the **browser's** cookie jar.
- The Capacitor WebView has a **separate** cookie jar.
- `rca://auth/callback` resumes the app — but the app's WebView still has no session cookie, so its next `GET /auth/me` returns 401 and the student is looking at the login screen again.

Resuming the app is not the same as being signed in inside it. Adding an `appUrlOpen` listener alone would not fix this; it would just make the failure quieter.

## What closing this needs

A one-time hand-off, roughly:

1. The callback detects a native-shell flow (a flag carried through `state`, so it cannot be forged from a query param).
2. Instead of setting the cookie and redirecting to a path, it mints a **single-use, short-TTL** hand-off code — a row in `portal_sessions`, or a separate table — and redirects to `rca://auth/callback?code=…`.
3. The app receives that via `App.addListener('appUrlOpen', …)` and `POST`s the code to something like `/auth/native/exchange`, **from the WebView**, which is what puts the cookie in the right jar.
[...]

PR #173, opened 2026-09-08
fix(portal): harden the OIDC verification path (closes #159, closes #160, closes #161, closes #162)

### LINKED ISSUE #159: [Portal] Add a nonce to the OIDC flow
Follow-up from the review on #127 (point 1), deferred from that PR.

For code+PKCE with a server-side exchange over TLS, replay is already blocked by `code_verifier`. A nonce closes the remaining token-substitution edge case — an ID token leaked from logs, a TLS-terminating proxy, or a provider bug getting injected here.

- Mint alongside `rca_oauth_state`
- Echo through as `?nonce=` in `start.js`
- Verify `claims.nonce === expected` in `verifyIdToken`

**Note:** @Pavankulli's #155 already implements this. If that PR becomes the OIDC foundation instead of #127, close this as already done.

### LINKED ISSUE #160: [Portal] Enforce the tenant in the issuer when MS_TENANT is pinned
Follow-up from the review on #127 (point 2), deferred from that PR.

`shared/oidc.js` accepts any issuer under `login.microsoftonline.com/`. That is correct for `common`, but wrong once someone sets `MS_TENANT=contoso` — a token from a *different* Entra tenant with a matching `client_id` would pass.

```js
if (!cfg.issuers) {
  const expected = env.MS_TENANT && env.MS_TENANT !== "common"
    ? new RegExp(`^https://login\\.microsoftonline\\.com/${env.MS_TENANT}/`)
    : /^https:\/\/login\.microsoftonline\.com\//;
  if (!expected.test(String(p.iss))) return null;
}
```

Needs `env` plumbed into `providerConfig` — a small refactor.

Not reachable today (nobody has pinned a tenant), which is why it was deferred. It becomes reachable the moment someone does.

### LINKED ISSUE #161: [Portal] Validate the session JWT header alg in verifySession
Follow-up from the review on #127 (point 3), deferred from that PR.

Decorative today — we are the only signer, HMAC-SHA256 output is length-fixed, and a bogus signature will not verify. It matters if a code path ever accepts asymmetrically-signed tokens through the same helper, at which point alg-confusion becomes reachable.

```js
const hdr = JSON.parse(dec.decode(b64urlToBytes(header)));
if (hdr.alg !== "HS256") return null;
```

Two lines, worth having before the helper grows a second caller.

### LINKED ISSUE #162: [Portal] Timeout and cache the JWKS fetch
Follow-up from the review on #127 (point 4), deferred from that PR.

`verifyIdToken` fetches JWKS with no abort signal, so if Google stalls, the callback hangs until Cloudflare's wall-clock cap — a login that hangs rather than fails.

- `AbortSignal.timeout(5000)` on the fetch
- A module-level cache respecting the response's `Cache-Control` (both providers set ~24h), which removes a round-trip from every login on the hot path

Both changes are in `shared/oidc.js`.

Abhigna1975:
The four follow-ups deferred from the review on #127. All of them live in code that has since landed on `main`, all four are unassigned, and none needed anyone else to unblock them.

Grouped into one PR because three touch the same twenty lines of `shared/oidc.js` — reviewing them apart would mean reading that file four times.

## #159 — nonce

`start.js` mints one alongside the state, sends it to the provider, and keeps it in an HttpOnly cookie. `callback.js` requires that cookie and hands the value to `verifyIdToken`, which rejects a mismatch.

code+PKCE over a server-side exchange already blocks replay of the *code*. The nonce closes token substitution — an ID token that leaked into a log or through a TLS-terminating proxy being injected at the callback. A token carrying **no** nonce when one is expected fails: a provider that drops it is one we cannot bind the flow to.

## #160 — tenant enforcement

The issuer check accepted anything under `login.microsoftonline.com/`, so with `MS_TENANT` pinned, a token from a *different* Entra tenant with a matching `client_id` passed.

**The fix suggested in the ticket would have broken logins**, and this is the part worth a close look. `MS_TENANT` may legitimately be a GUID, a verified domain (`contoso.onmicrosoft.com`), or `common`/`organizations`/`consumers` — but `iss` always carries the tenant **GUID**. Regex-matching a domain value against the issuer would reject every valid sign-in.

So `isTrustedMicrosoftIssuer` tightens only when the pin is a GUID — the case we can be certain about — and leaves a domain pin on the host check, with a comment saying why. Anything not under Microsoft's host is refused either way.

## #161 — algorithm pinning

`verifySession` parses the header and requires `HS256` before touching the token.

To be clear about severity: **this is not a live vulnerability.** An attacker cannot produce a validly-signed token without the secret, and the ticket says as much. It is defence in depth for the moment a second caller hands this helper an asymmetrically-signed token — which #163 is about to create.

## #162 — JWKS timeout and cache

The fetch had no abort signal, so a stalled provider hung the callback until Cloudflare's wall-clock cap — a login that hangs rather than fails. Now `AbortSignal.timeout(5000)`, plus a module-scope cache honouring the response's `max-age`, clamped to 5 minutes .. 24 hours so neither `max-age=0` nor a year is taken literally.

Two details that matter more than the caching:

- **A cached set without the token's `kid` is stale by definition, so it refetches.** Without that, a provider key rotation would fail every login until the isolate happened to be evicted.
- **If the refetch fails, an expired cache is still used.** Refusing every login because the provider is briefly unreachable is worse than trusting public keys a few hours past their max-age.

## Tests

21 new, in `tests/portal.oidc-hardening.test.js`. They generate a real RSA keypair and sign real RS256 ID tokens, so the verification path is genuinely exercised rather than stubbed. Suite goes **177 → 198**.

**14 of the 21 fail on `main`** — I ran them there rather than assuming.

The #161 tests are built the awkward way on purpose. Swapping a header and keeping the original signature proves nothing: the HMAC covers the header, so the signature check rejects it with or without the fix, and that test passes either way. They sign the swapped header correctly with the secret instead — which is what an attacker holds in a real alg-confusion — so they actually fail on `main`. There is a control test alongside them establishing that the helper produces something the verifier would otherwise accept.

Build clean, lint clean on every touched file.

### DIFF
--- a/functions/auth/[provider]/callback.js
+++ b/functions/auth/[provider]/callback.js
@@ -4,7 +4,7 @@ import { signSession, sessionCookie, readCookie } from "../../../shared/auth.js"
 import { isConfigured, providerConfig, redirectUri, verifyIdToken } from "../../../shared/oidc.js";
 
 const clearTx = (headers) => {
-  for (const n of ["rca_oauth_state", "rca_oauth_verifier", "rca_oauth_next"]) {
+  for (const n of ["rca_oauth_state", "rca_oauth_verifier", "rca_oauth_nonce", "rca_oauth_next"]) {
     headers.append("set-cookie", `${n}=; HttpOnly; Secure; SameSite=Lax; Path=/auth; Max-Age=0`);
   }
 };
@@ -57,7 +57,12 @@ export async function onRequestGet(context) {
   }
   if (!tokens.id_token) return fail("token_exchange");
 
-  const claims = await verifyIdToken(tokens.id_token, cfg);
+  // The nonce cookie is set by our own start endpoint, so a request that
+  // arrives without one did not begin here (#159).
+  const expectedNonce = readCookie(request, "rca_oauth_nonce");
+  if (!expectedNonce) return fail("bad_state");
+
+  const claims = await verifyIdToken(tokens.id_token, cfg, fetch, expectedNonce);
   if (!claims) return fail("bad_token");
 
   if (!env.DB) return fail("storage");
--- a/functions/auth/[provider]/start.js
+++ b/functions/auth/[provider]/start.js
@@ -27,6 +27,10 @@ export async function onRequestGet(context) {
   const state = randomString(16);
   const verifier = randomString(32);
   const challenge = await codeChallenge(verifier);
+  // Binds the ID token to this browser (#159). PKCE already blocks replay of
+  // the code; the nonce is what stops a token minted elsewhere being swapped
+  // in at the callback.
+  const nonce = randomString(16);
 
   // Where to land inside the app afterwards. Only a same-site path is kept —
   // an absolute URL here would make this an open redirect.
@@ -41,6 +45,7 @@ export async function onRequestGet(context) {
   url.searchParams.set("state", state);
   url.searchParams.set("code_challenge", challenge);
   url.searchParams.set("code_challenge_method", "S256");
+  url.searchParams.set("nonce", nonce);
   // Ask for an account chooser rather than silently reusing whichever account
   // the phone happens to be signed into.
   url.searchParams.set("prompt", "select_account");
@@ -48,6 +53,7 @@ export async function onRequestGet(context) {
   const headers = new Headers({ location: url.toString() });
   headers.append("set-cookie", txCookie("rca_oauth_state", state));
   headers.append("set-cookie", txCookie("rca_oauth_verifier", verifier));
+  headers.append("set-cookie", txCookie("rca_oauth_nonce", nonce));
   headers.append("set-cookie", txCookie("rca_oauth_next", encodeURIComponent(next)));
   return new Response(null, { status: 302, headers });
 }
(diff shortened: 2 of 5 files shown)


nikshepkulli (approved):
LGTM. All four follow-ups from the #127 review landed exactly as intended:

- **#159 nonce binding** — required at callback (drop-nonce provider = fail, not a silent pass). The "PKCE blocks replay of the code; nonce stops a token minted elsewhere being swapped in" comment is the correct framing.
- **#160 MS tenant pin** — GUID-only tightening avoids rejecting valid logins when `MS_TENANT` is a domain or one of the meta-values (`common` / `organizations` / `consumers`). Right call to leave the domain case on the host check rather than adding a discovery round trip.
- **#161 alg pinning** — the "unreachable today, stops being decorative the moment a second caller hands this helper an asymmetrically-signed token" is exactly the reason to add it now. Alg-confusion is cheap to prevent and expensive to detect.
- **#162 JWKS cache** — kid-aware refetch handles rotation without waiting for isolate eviction. The `max-age` clamp (300s min, 24h max) is a good hedge against a provider sending `max-age=0` or `max-age=1y`. Stale-on-network-failure is the right failure mode — refusing every login while Google is unreachable is worse than serving one-day-old keys.

235-line test file covers all four. Approving. Ready to merge.

Issue #186, opened 2026-09-10
[Portal] Phase 2 — Instructor upload API: lesson video to R2 + lesson create/edit/publish


Abhigna1975:
Part of #43 (epic #41). Needs the R2 binding and the role guard.

- Resumable (multipart) upload of a lesson's video to R2; on completion set `lessons.video_url` and `duration_seconds`.
- Create, edit, reorder and publish lessons for a course.
- Instructor/admin only.

**Open question on #43:** R2 doesn't transcode, so the low-res copy is either the instructor's 360p export or a transcoding step.

**Done when:** an instructor can upload a video to a lesson and publish it, and a student then sees it in the course API.

Issue #184, opened 2026-09-10
[Portal] Phase 2 — R2 bucket + binding for lesson video


Abhigna1975:
Part of #43 (epic #41).

- Create an R2 bucket for lesson video and bind it in `wrangler.toml` (e.g. `LESSONS`), next to the D1 `DB` binding.
- Document the bucket name and any CORS the Android app needs (the shell serves from `https://localhost`, #163).

**Done when:** a Pages Function can read and write an object in the bucket on preview and production.

PR #198, opened 2026-09-15
feat(portal): video stream API with HTTP range requests (#187)

### LINKED ISSUE #187: [Portal] Phase 2 — Video API: stream a lesson from R2 with range requests
Part of #43 (epic #41). Needs the R2 binding.

- `GET /api/portal/lessons/:id/video`: session + enrolment check (404 when not enrolled, as `/api/portal/courses/:slug` does), served from R2.
- HTTP range requests, so the player can seek and resume, and a download can pick up where a dropped connection left it.
- CORS for the Android shell, as the course APIs do (#163).

**Done when:** an enrolled student can stream and seek a lesson; anyone else gets 404.

Pavankulli:
## What

Adds `GET /api/portal/lessons/:lessonId/video` — streams a lesson video from R2 with full HTTP range-request support so the player can seek and a download can resume after a dropped connection.

## Access control

Matches the `/api/portal/courses/:slug` pattern exactly:
- Student must have a valid session
- Student must be enrolled in the course that owns the lesson
- Both the course and the lesson must be published
- Every failure returns **404** — no distinction between "not enrolled" and "doesn't exist" so enrolment state cannot be probed by guessing lesson IDs

## Response behaviour

| Condition | Status |
|-----------|--------|
| No session | 401 |
| DB fault | 503 |
| LESSONS R2 binding missing | 503 |
| Not enrolled / unpublished / no video yet / R2 object deleted | 404 |
| Malformed or inverted `Range` header | 416 |
| No `Range` header | **200** + `Accept-Ranges: bytes` + `Content-Length` |
| Valid `Range` header | **206** + `Content-Range` + `Content-Length` |

## Range forms handled

```
Range: bytes=0-999      → { offset: 0, length: 1000 }
Range: bytes=500-       → { offset: 500 }          (open-ended to EOF)
Range: bytes=-500       → { suffix: 500 }           (last 500 bytes)
```

`Content-Range` and `Content-Length` are derived from `r2Object.range` (the actual range R2 served), so partial reads at end-of-file report correctly without an extra HEAD round-trip.

## Shared change

`shared/courses.js` — `getLessonForStream(db, userId, lessonId)`: 3-table JOIN (lessons + courses + enrolments_users) that verifies all access conditions atomically.

## Tests

`tests/portal.video.test.js` — 20 tests covering:
- `getLessonForStream`: happy path, null on unenrolled, bind args, published filters verified in SQL
- Endpoint auth guards (401, 503, 404 for each failure mode)
- 200 full response headers
- 206 for `bytes=X-Y`, `bytes=X-`, `bytes=-N`
- 416 for malformed and inverted Range
- OPTIONS preflight (204 native origin, 403 other)

**403/403 tests passing.**

## Dependencies

- Requires #184 (LESSONS R2 binding) — included via branch ancestry from #186/#185/#184

Closes #187.


Pavankulli:
## Self-review — 6 bugs fixed (commit 3dfb163)

### Bugs found and fixed

| # | Location | Bug | Fix |
|---|----------|-----|-----|
| 1 | `parseRangeHeader` | `bytes=-0` returned `{ suffix: 0 }` — not satisfiable per RFC 7233 | Return `false` when `n === 0` → 416 |
| 2 | `parseRangeHeader` | Multi-range `bytes=0-100,200-300` returned `null` → triggered 416; RFC 7233 §2.1 says server MAY ignore and serve 200 | Detect comma → return `null` (serve full 200) |
| 3 | `onRequestGet` | `rangeHeader && !rangeOption` caught both malformed AND multi-range, both → 416 | Distinguish: `false` → 416, `null` → full 200 |
| 4 | `onRequestGet` | Open-ended `bytes=X-` never validated against file size; R2 would receive out-of-bounds offset | Post-fetch check: `rangeOption.offset >= object.size` → 416 with `bytes */{size}` |
| 5 | `onRequestGet` | Suffix range: `object.range` fallback `start=0, length=totalSize` sent full file with wrong Content-Range | Derive `start/length` from `rangeOption.suffix` when `object.range` absent |
| 6 | `onRequestGet` | Content-Type hardcoded `"video/mp4"` — won't generalise if storage type changes | Use `object.httpMetadata?.contentType ?? "video/mp4"` |

### Tests added (5 new, 24 total)
- `bytes=-0` → 416
- open-ended offset >= file size → 416 + correct `Content-Range: bytes */10000000`
- multi-range → 200 (full content)
- single-byte `bytes=0-0` → 206 + `Content-Range: bytes 0-0/10000000`
- suffix `bytes=-500` now asserts `Content-Range` and `Content-Length` (was false positive before fakeR2 suffix fix)

### fakeR2 fix
`fakeR2` didn't handle `{ suffix }` option — suffix tests were passing vacuously. Fixed: `offset = size - suffix`, `length = suffix`.

